PR #2, opened 2018-10-17
Add securedrop-workstation-grsec metapackage

emkll:
This metapackage will track the latest available kernel and headers
version for the securedrop-workstation qubes template. It will also
include libelf-dev used for installing the Qubes kernel module, as well
as paxctld to manage PaX flags on binaries.

(copied from existing logic in
https://github.com/freedomofpress/securedrop-workstation)
Bump kernels to 4.14.74

Fixes https://github.com/freedomofpress/securedrop-workstation/issues/165

This package has already been uploaded to qubes-apt-test.freedom.press

Testing instructions:
* `cd securedrop-workstation-grsec`
* `dpkg-buildpackage -us -uc`
* `dpkg -I securedrop-workstation-grsec_4.14.74-1_amd64.deb` returns the correct kernel versions in the `Depends` field.


securedrop-workstation-grsec/debian/postinst
```diff
@@ -0,0 +1,44 @@
...
+# for details, see https://www.debian.org/doc/debian-policy/ or
+# the debian-policy package
+
+
+case "$1" in
+    configure)
+    # DKMS autoinstall the qubes kernel modules
+    dkms autoinstall 4.14.74-grsec
+    # restart paxctld to set pax flags (incl grub)
+    service paxctld restart
```
conorsch: `service paxctld restart` uses the old SysV-style init syntax. After running this command manually, the paxctld service is indeed started, but it is not marked as "enabled"—meaning it won't start on a subsequent boot. Let's change this line to:

```
systemd enable paxctld
systemd restart paxctld
```

Then it should persist will on subsequent boots.

PR #5, opened 2018-11-01
Adds the initial wheels urls and sha256sums

kushaldas:
We also have the script to create the wheelurls.txt file.

You can download all the files to ./localwheels directory using the following commands.

```
mkdir localwheels
wget -i wheelsurls.txt -P ./localwheels/
```

Update the sha256sums.txt file manually if required by running in the `sha256sum *` in any `localwheels` directory.

To create the `wheelsurls.txt`:

```
./scripts/createdownloadurls.py > wheelsurls.txt
```


scripts/createdownloadurls.py
```diff
@@ -0,0 +1,25 @@
...
+        words = line.split()
+        if len(words) == 2:
+            filenames.append(words[1])
+        else:
+            print("Worng line: {0}".format(line))
+
+    filenames = list(set(filenames))
+    filenames.sort()
+    for name in filenames:
+        print("https://dev-bin.ops.securedrop.org/localwheels/{0}".format(name))
```
msheiny: can we provide an over-ride here (maybe env var with a default) on the base URL? 

`dev-bin.ops.securedrop.org` is for testing/development... im planning to promote wheels that are good to `bin.ops.secuerdrop.org`

scripts/createdownloadurls.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+
```
conorsch: Add a header comment here to explain what this script does.

scripts/createdownloadurls.py
```diff
@@ -0,0 +1,25 @@
...
+    filenames = []
+    for line in lines:
+        line = line.strip()
+        if not line:
+            continue
+        words = line.split()
+        if len(words) == 2:
+            filenames.append(words[1])
+        else:
+            print("Worng line: {0}".format(line))
```
conorsch: Nit: typo `s/Worng/Wrong/`

scripts/createdownloadurls.py
conorsch: As @msheiny noted, making this overridable is wise. Something like:

```
DEV_WHEELS_BUCKET_BASE_URL = "https://dev-bin.ops.securedrop.org/localwheels"
WHEELS_BUCKET_BASE_URL = os.environ.get("WHEELS_BUCKET_BASE_URL", DEV_WHEELS_BUCKET_BASE_URL)
```

at the top of the file.

kushaldas: @msheiny Tell me what name you want for the envvar?

scripts/createdownloadurls.py
kushaldas: @conorsch I used `black` in the latest push.

conorsch:
LGTM @kushaldas, thanks for the quick fixes!

PR #9, opened 2018-11-02
Makefile targets and scripts and also the FPF PyPI index

kushaldas:
Right now updating the FPF PyPI index is a bit manual step. But will auomate those later.

Needs the following PRs to be merged to test successfully.

- [x] https://github.com/freedomofpress/securedrop-client/pull/101
- [x] https://github.com/freedomofpress/securedrop-proxy/pull/16

If you want to test before those get merged, just do the following two steps in proxy and client repo.

```
pipenv lock -r > requirements.txt
Remove the top line of requirements.txt which points to pypi.org
```

After that, run `python3 setup.py sdist` in both `securedrop-client` or `securedrop-proxy` directory.
Remember the `dist/securedrop-NAME-version.tar.gz` file path. and also the VERSION.

### How to test?

-  Install dependencies `make install-deps`
-  Sync wheels from top (just to check) `make syncwheels`
- To build `securedrop-client`, I used this command. Modify for your path to the tarball. Similarly try securedrop-proxy (version is `0.1.0` there)

```
PKG_PATH=/home/user/code/securedrop-client/dist/securedrop-client-0.0.1.tar.gz PKG_VERSION=0.0.1 make securedrop-client
```


conorsch:
We're going to be revisiting the packaging story, with a strong eye toward integrity, over the coming days. The PyPI index generation logic could be tweaked to retrieve URLs that include the hashes of the wheel files—hat tip to @emkll for that suggestion.

PR #10, opened 2018-11-06
Creates securedrop-workstation-svs-disp package

emkll:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/182

This provides the sd-svs-disp template that will provide the applications to view certain file types in disposable VMs, as well as manage PaX flags for the template.

PaX flags are default for paxctld, with the sole addition of the `m` flag for `/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java`.

I have considered adding pax flags via securedrop-grsec metapackage, but the squash logic might become problematic if we are to override PaX flags on a per-vm basis.

Once this is merged, I will upload package to apt-test-qubes and open a PR in securedrop-workstation to install the package in the template as part of the provisioning process.

### Test plan

- [ ] Build both securedrop-workstation-grsec and securedrop-workstation-svs-disp debian packages by entering the directory and running `dpkg-buildpackage -uc -us`.
- [ ] Install both packages inside `sd-svs-disp` Template VM. You will need to run `apt-get -f install` afterwards to get dependencies to resolve since this is a local install and not via apt.
- [ ] In `sd-svs-disp`
- [ ] Mimetypes are properly respected (e.g. .doc file opens correctly, plaintext files are also opened properly)


conorsch:
> I have considered adding pax flags via securedrop-grsec metapackage, but the squash logic might become problematic if we are to override PaX flags on a per-vm basis.

Bundling the pax flags with the securedrop-grsec metapackage seems the best approach, for a few reasons:

* Dependency ordering: setting the securedrop-grsec metapackage as a dependency of a given AppVM pkg ensures that the relevant PaX flags are configured _before_ the applications are installed, eliminating a window when applications will not work correctly during updates. (Given the Template/AppVM setup, this is not critical from a usability perspective, but the PaX flags may be required in order for e.g. postinst hooks to complete.)
* Simplicity: we should manage a single list of PaX flags that gets installed in _all_ VMs; if a given binary is not present, paxctld will simply ignore the declaration

Just noting those points for discussion; I've yet to step through the test plan as described.

msheiny:
> We're still having issues with a race condition during java install, where paxctl doesn't fix the flags fast enough for the postinst scripts to complete. For now, let's simply remove the libreoffice dependency (since that's what pulls in java), and we can proceed with packaging and testing the other filetype handlers.

Hey! I made this comment out of band in gitter. The only way i was able to fix this in the past was...

- modify pax config
- attempt to install java tooling (it'll fail)
- restart pax
- second round of installing java tooling (some of the files are already laid down and itll try to do the postinst stuff again)

conorsch:
Thanks for chiming in, @msheiny. The strategy you suggest would work well in containers, or even via a config management tool, but it's not an option during postinst, due to dpkg locks. If we absolutely must, we could move the libreoffice install logic to Salt (rather than apt dependencies). 

Another option @emkll suggested is to bake it into a template that gets installed into dom0—either the existing [qubes-template-securedrop-workstation](https://github.com/freedomofpress/qubes-template-securedrop-workstation) or some derivative thereof. Neither option is particularly attractive—so we'll definitely be having this discussion again. :slightly_frowning_face:

emkll:
Thanks @conorsch for the review, I've removed LibreOffice dependency for now to unblock the provisioning logic, and we can revisit later. Bumped the version because 0.1.0 was already uploaded to apt server.

The package built from this PR (securedrop-workstation-svs-disp 0.1.1) is now on apt-test-qubes.freedom.press

conorsch:
Approving base on visual review. We've already got the packages served, so next step is to start on the integration logic. We can bump versions over here as we uncover problems.

PR #21, opened 2018-11-12
Removes dch and localwheels syncing part of build steps

### LINKED ISSUE #19: Remove dch call from the building package script
If we have the `dch` command in the `build-debianpackage` script, it will create a lot of confusion in a longer run where we have to build and test packages regular (without releasing on the repos).

For example, I just created a set tarball with the same version and wanted to rebuild the package. This is what I got from the current tooling.

double_changelog

We should remove this step from the build script and instead add a check for the final debian packages (also in the CI in future). This is a human problem (the updating of the changelogs in the correct way), and we should not try to automate away the problem (as it will create more problem in a longer run).

kushaldas:
Fixes #19. 

### How to test?

Build of any of the existing package. Remember to update the path based on the location of the source tarball.

```
PKG_PATH=/home/user/code/securedrop-client/dist/securedrop-client-0.0.5.tar.gz PKG_VERSION=0.0.5 make securedrop-client
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,11 +9,11 @@ syncwheels: ## Downloads wheels and sources from the remote server
 	./scripts/syncwheels
 
 .PHONY: securedrop-proxy
-securedrop-proxy: syncwheels ## Builds Debian package for securedrop-proxy code
+securedrop-proxy: ## Builds Debian package for securedrop-proxy code
 	PKG_NAME="securedrop-proxy" ./scripts/build-debianpackage
 
 .PHONY: securedrop-client
-securedrop-client: syncwheels ## Builds Debian package for securedrop-client code
+securedrop-client: ## Builds Debian package for securedrop-client code
 	PKG_NAME="securedrop-client" ./scripts/build-debianpackage
 
 .PHONY: securedrop-workstation-config
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -51,10 +51,6 @@ mkdir -p "${TOP_BUILDDIR}/${PKG_NAME}"
 
 printf "Building package '%s' from version '%s'...\\n" "$PKG_NAME" "$PKG_VERSION"
 
-# Update changelog interactively. Currently using "unstable" but we intend
-# to target "stretch" for SDW applications. Must run before files are copied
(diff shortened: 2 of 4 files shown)


conorsch:
Rebasing on latest master (5411e3f6762d0fe94b49be9fab6bce93c90469be) for review.

scripts/build-debianpackage
```diff
@@ -51,10 +51,6 @@ mkdir -p "${TOP_BUILDDIR}/${PKG_NAME}"
 
 printf "Building package '%s' from version '%s'...\\n" "$PKG_NAME" "$PKG_VERSION"
 
-# Update changelog interactively. Currently using "unstable" but we intend
-# to target "stretch" for SDW applications. Must run before files are copied
-# into non-version-controlled build dir.
-dch --newversion "${PKG_VERSION}" --distribution unstable -c "${CUR_DIR}/${PKG_NAME}/debian/changelog"
```
conorsch: The reason we added this auto-dch command is two-fold:

1. It's extremely easy to overlook doing it yourself, thus the script; but we can learn to run a command ad-hoc, it won't kill us.
2. There are a bunch of flags that should be added to the `dch` invocation, particularly the `--distribution`, that's quite painful to recall from memory on every run.

I'll read the manpage for dch completely before suggesting any changes; maybe tossing in a `scripts/dch-wrapper` would accommodate the manual workflow with flags...

Makefile
```diff
@@ -9,11 +9,11 @@ syncwheels: ## Downloads wheels and sources from the remote server
 	./scripts/syncwheels
 
 .PHONY: securedrop-proxy
-securedrop-proxy: syncwheels ## Builds Debian package for securedrop-proxy code
+securedrop-proxy: ## Builds Debian package for securedrop-proxy code
 	PKG_NAME="securedrop-proxy" ./scripts/build-debianpackage
 
 .PHONY: securedrop-client
-securedrop-client: syncwheels ## Builds Debian package for securedrop-client code
```
conorsch: The `syncwheel` actions were removed because we actually want to verify the hashes in the repo vc, and trust pip to fail if the hashes are incorrect when fetching from the PyPI mirror.

PR #23, opened 2018-11-12
Makefile updates and tools for requirements.txt and wheels

kushaldas:
This PR has a few new scripts and related Makefile update. @conorsch I am guessing you will update the Makefile bit to make a bit more professional :)

These scripts and Makefile targets will help our package maintainers to create `requirements.txt` and `requirements-dev.txt` files for the actual release. Those files will contain the sha256sums for the binary wheels from our PyPI (the s3 bucket).

### How to test?

First, let us the create/update the files, requires `PKG_DIR` environment variable.

```
PKG_DIR=/home/user/code/securedrop-client make requirements
```

This command will update the requirements/versions from the upstream PyPI. If there is any new update or if any of the new binary wheels are missing, this will fail. It will also print the next command a developer can use to generate the missing wheels.

```
$ PKG_DIR=/home/user/code/securedrop-client ./scripts/update-requirements 
The following dependent wheel(s) are missing:
pytest==3.10.1

Please build the wheel by using the following command:

	PKG_DIR=/home/user/code/securedrop-client make build-wheels

Then sync the newly built wheels and sources to the s3 bucket.
Also update the index HTML files accordingly and sync to s3.
After these steps, please rerun the command again.
```
If such missing wheels are there (you can delete any related wheel and test this), the above mentioned command will create such new wheels.

```
PKG_DIR=/home/user/code/securedrop-client make build-wheels
```

After this, remember to sync the `localwheels` directory to the s3, and also update the HTML index files and also the `sha256sums.txt` files as required. Also remember to deploy any new wheel+index to the s3 bucket.

The `sha256sums.txt` should always get new entries, but, there should not be any update of the any existing wheels or hashes. If there is any such change, it means one did a rebuild of the wheel which is already there in our s3/PyPI index. So, from the security aspect, we will have to check that in the CI for each commit.


kushaldas:
Please remember that the `pytest-3.10.1` is an example, I have already built and uploaded that to the s3.

conorsch:
Rebasing on latest master (5411e3f6762d0fe94b49be9fab6bce93c90469be) to review.

Makefile
```diff
@@ -26,6 +26,23 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
...
 
+.PHONY: requirements
+requirements: ## Creates requirements files for the Python projects
+	./scripts/create-requirements
+	./scripts/update-requirements
+
+.PHONY: build-wheels
+build-wheels: syncwheels ## Builds the wheels and syncs to the localwheels directory
+	./scripts/build-sync-wheels -r ${PKG_DIR}/requirements.txt
+	./scripts/build-sync-wheels -r ${PKG_DIR}/requirements-dev.txt
```
redshiftzero: we don't need this right?

Makefile
```diff
@@ -26,6 +26,23 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
...
 	./scripts/install-deps
 
+.PHONY: requirements
+requirements: ## Creates requirements files for the Python projects
+	./scripts/create-requirements
+	./scripts/update-requirements
+
+.PHONY: build-wheels
+build-wheels: syncwheels ## Builds the wheels and syncs to the localwheels directory
+	./scripts/build-sync-wheels -r ${PKG_DIR}/requirements.txt
```
redshiftzero: so we don't need the requirements*.txt files in any of the proxy or client repos if this is merged, correct?

scripts/create-requirements
```diff
@@ -0,0 +1,19 @@
...
+if [[ -z "${PKG_DIR:-}" ]]; then
+    echo "Set PKG_DIR of the project.";
+    exit 1
+fi
+
+cd $PKG_DIR
+
+# Now create both of the files
+pipenv lock -r > requirements.txt
+pipenv lock -d -r > requirements-dev.txt
```
redshiftzero: remove this line?

scripts/sync-sha256sums
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# A script to update the sha256sums from localwheels directory
+
+set -e
+set -u
+set -o pipefail
+
+cd ./localwheels/
+sha256sum * > ../sha256sums.txt
```
redshiftzero: let's document the fact that this file should be signed to the README

scripts/update-requirements
```diff
@@ -0,0 +1,133 @@
...
+    req_path = os.path.join(PKG_DIR, "requirements.txt")
+    reqdev_path = os.path.join(PKG_DIR, "requirements-dev.txt")
+    for path in [req_path, reqdev_path]:
+        if not os.path.exists(path):
+            print("Missing {0}.".format(path))
+            sys.exit(1)
+
+    # First remove index line and any PyQt or sip dependency
+    cleanup(req_path)
+    cleanup(reqdev_path)
```
redshiftzero: remove

scripts/update-requirements
```diff
@@ -0,0 +1,133 @@
...
+
+    # Now let us update the files along with the sha256sums from localwheels
+    add_sha256sums(req_path)
+    add_sha256sums(reqdev_path)
```
redshiftzero: remove

scripts/update-requirements
```diff
@@ -0,0 +1,133 @@
...
+    # Do not update the file if missing wheels
+    if missing_wheels:
+        print("The following dependent wheel(s) are missing:")
+        for value in missing_wheels:
+            print("{0}=={1}".format(value[0], value[1]))
+
+        print("\nPlease build the wheel by using the following command:\n")
+        print("\tPKG_DIR={0} make build-wheels\n".format(os.environ["PKG_DIR"]))
+        print("Then sync the newly built wheels and sources to the s3 bucket.")
+        print("Also update the index HTML files accordingly and sync to s3.")
```
redshiftzero: Let's link in the README this issue: https://github.com/freedomofpress/securedrop-debian-packaging-guide/issues/6 

otherwise this is buried and is difficult for people to find

conorsch:
> After this, remember to sync the localwheels directory to the s3, and also update the HTML index files and also the sha256sums.txt files as required. Also remember to deploy any new wheel+index to the s3 bucket.

Those steps should absolutely be automated—the "syncwheels" terminology presently in the repo should perhaps be called "fetch-wheels", since it doesn't actually synchronize them. Detailed instructions for syncing the wheels can be found over in https://github.com/freedomofpress/securedrop-debian-packaging-guide/issues/6 ; those AWS S3 commands should absolutely be pulled into this repo's README, and ideally tooling, in order to bridge gaps in the dev workflow.

Makefile
redshiftzero: or do we use this script to generate those files and then commit them in the upstream repository? we should document this in the README if so

scripts/update-requirements
```diff
@@ -0,0 +1,133 @@
...
+    :rtype: None
+    """
+
+    with open(path) as fobj:
+        lines = fobj.readlines()
+
+    finallines = []
+
+    for line in lines:
+        # Skip the PyPI index line
```
redshiftzero: `# Skip the PyPI index line` -> `# Skip the PyPI index line so that when we install we use the FPF PyPI mirror`

scripts/update-requirements
```diff
@@ -0,0 +1,133 @@
...
+        if line.startswith("-i https://pypi.org/simple"):
+            continue
+        # For PyQt5
+        elif line.startswith("pyqt5"):
+            continue
+        # For sip
+        elif line.startswith("sip"):
```
redshiftzero: why? (needs comment)

scripts/update-requirements
```diff
@@ -0,0 +1,133 @@
...
+        lines = fobj.readlines()
+
+    finallines = []
+
+    for line in lines:
+        # Skip the PyPI index line
+        if line.startswith("-i https://pypi.org/simple"):
+            continue
+        # For PyQt5
+        elif line.startswith("pyqt5"):
```
redshiftzero: why? (note: I know why but this is something we should document in the script for the benefit of future us)

scripts/create-requirements
conorsch: Having the requirements.txt files generated (and later committed) _in a separate repo_ strikes me as a bad pattern. Let's instead pull those generated reqs files into _this_ repo, and commit them here. 

Those reqs files also must have hashes included (from the Pipfile.lock is fine).

redshiftzero:
So the python wheel builder steps are: 

1. `make requirements` -> Generates requirements files for each project

2. `make sync-sha256sums` -> Generates SHA256 hashes for the tarballs 

Important to add as discussed this morning: This should then bail if the SHA256 hashes (of the tarballs only) are not present what is in Pipfile.lock in the package

3. `make build-wheels` -> Builds wheels

Important to add: 
 * Whenever SHA256sums are edited, the packager must be prompted to sign the document (in the script and in the README)
 * Whenever SHA256sums are used, the signature must be verified. If it does not verify, if it is not signed with the proper key, or if it does not exist, the Makefile target should bail. 

These steps, if you agree (if not please correct them when you add to the README) we should add to the README.

redshiftzero:
The other item we need to add is logic to scripts/build-debianpackage to bail if the signature on the document that contains the SHA256sums of the built wheels does not verify

Makefile
kushaldas: >we don't need this right?

The build-wheels action is to help us build all the dependencies into wheels. We do need this.

Makefile
kushaldas: requirements*.txt files will be stored in the git and also be part of the source release. The pip will use these files to do the final install for the debian package. I will update the guide and readme for this.

scripts/create-requirements
kushaldas: We are keeping both of these files.

scripts/update-requirements
kushaldas: > Those steps should absolutely be automated—the "syncwheels" terminology presently in the repo should perhaps be called "fetch-wheels", since it doesn't actually synchronize them. Detailed instructions for syncing the wheels can be found over in [freedomofpress/securedrop-debian-packaging-guide#6](https://github.com/freedomofpress/securedrop-debian-packaging-guide/issues/6) ; those AWS S3 commands should absolutely be pulled into this repo's README, and ideally tooling, in order to bridge gaps in the dev workflow.

There should more/better commands to do this, but the actual sync to s3 + signing the doc etc should be done manually as it helps to verify what exactly we are syncing.

Makefile
redshiftzero: sorry for lack of clarity, i meant we don't need the dev requirements files specifically, since we're not using them in the build

kushaldas: >sorry for lack of clarity, i meant we don't need the dev requirements files specifically, since we're not using them in the build

Yup, and I removed that file. I forgot to update the comment here :(

Makefile
```diff
@@ -26,6 +26,22 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
 install-deps: ## Install initial Debian packaging dependencies
 	./scripts/install-deps
 
+.PHONY: requirements
+requirements: ## Creates requirements files for the Python projects
+	./scripts/create-requirements
+	./scripts/update-requirements
```
redshiftzero: This makefile target which updates the requirements file in an adjacent project assumes several things: 

1. The user has cloned this debian packaging repo
2. The user can upload new requirements to our PyPI mirror
3. The user has the keys of the signer of the sha256sums.txt in their gpg keyring (FPF staff do, others do not)

I don't think it's reasonable that a random developer adding a new requirement should need to do this to update `requirements.txt` in the project repository. This seems to me reasonable to be part of either the release manager for the project or the packager's responsibility. 

I can live with the release manager doing the following: 

1. Bumps versions where needed
2. Does a test build (or delegates someone to do so) 
3. Updates requirements file via `make requirements` after setting the `PKG_DIR` of the project - we add a note perhaps in the top of the file indicating that these are for the build process only so that developers don't get confused (it is definitely unclear to have duplicated requirements files, one in Pipfile and one called `requirements.txt`)
4. They may or may not need to build and push new wheels to the FPF PyPI 
5. Finally, make release PR and generate tag to build from

Makefile
```diff
@@ -26,6 +26,22 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
...
+
+.PHONY: build-wheels
+build-wheels: syncwheels ## Builds the wheels and syncs to the localwheels directory
+	./scripts/build-sync-wheels -p ${PKG_DIR}
+	./scripts/sync-sha256sums
+	./scripts/createdownloadurls.py > wheelsurls.txt
+
+.PHONY: sync-sha256sums
```
redshiftzero: Looks like we _don't_ need this makefile target? If it is not required, let's remove it, or document in what scenario it should be used.

Makefile
```diff
@@ -26,6 +26,22 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
...
 
+.PHONY: requirements
+requirements: ## Creates requirements files for the Python projects
+	./scripts/create-requirements
+	./scripts/update-requirements
+
+.PHONY: build-wheels
+build-wheels: syncwheels ## Builds the wheels and syncs to the localwheels directory
+	./scripts/build-sync-wheels -p ${PKG_DIR}
+	./scripts/sync-sha256sums
```
redshiftzero: At this stage we've updated `sha256sums.txt`, but we need to include either in this makefile target or in the documentation a note that we need to sign this document again.

Makefile
```diff
@@ -26,6 +26,22 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
...
+	./scripts/createdownloadurls.py > wheelsurls.txt
```
redshiftzero: Finally at this stage we can push the modified wheels and updated PyPI index files to S3, so we can again add a note (just print to the user "run aws ...")

scripts/update-requirements
redshiftzero: looks like this requirement is used in dev only, but doesn't hurt to keep in here

Makefile
conorsch: The dev-facing instructions in `scripts/update-requirements` recommend its use `if missing_wheels`; so looks like it should stay for now.

redshiftzero: isn't that `make build-wheels` not `make sync-sha256sums`?

redshiftzero: ok implemented

Makefile
```diff
@@ -26,6 +26,22 @@ securedrop-workstation-grsec: ## Builds Debian metapackage for Qubes Workstation
 install-deps: ## Install initial Debian packaging dependencies
 	./scripts/install-deps
 
+.PHONY: requirements
+requirements: ## Creates requirements files for the Python projects
+	./scripts/create-requirements
+	./scripts/update-requirements
+
+.PHONY: build-wheels
+build-wheels: syncwheels ## Builds the wheels and syncs to the localwheels directory
```
redshiftzero: at this step we need to dump the current shasums from what we just downloaded from S3 to a new sha256sums-downloaded.txt file, and then verify the sig on the existing sha256sums.txt, and diff the two files. Or we can end up signing the hash of a compromised wheel from S3

Makefile
redshiftzero: gonna update README in this repository indicating as such

redshiftzero: (note that if we can come up with a story that isn't terrible from the perspective of potential contributors without access to all this internal FPF tooling then I'm open to it, otherwise we need to live with this workflow)

scripts/update-requirements
redshiftzero: I've done this in the code now

PR #33, opened 2018-12-04
Fixes #32 brings reproducible builds back

### LINKED ISSUE #32: Justify or remove wheel build process
In the last engineering meeting, we discussed potential simplifications to the debian package build process in this repository. One idea is to abandon the hosting of built wheels, and instead build the wheels from the tarballs from PyPI at debian package build time. Right now, this doesn't gain us much, since the debian package build process is not reproducible. If we were to remove this step, it would enable us to remove our custom PyPI index and tooling required to generate/update it.

However, one reason to keep the wheel build process is such that one we have built wheels, we can have a reproducible deb build process. @kushaldas I believe you took a look at this - what remaining steps need to be done for reproducible debian package builds?

kushaldas:
Using these changes we can have reproducible builds, as we will be using the same source tarball and also the same wheels to build the final package.

Fixes #32.

### How to test?

Build the `securedrop-client` twice, after the first build, move the deb package to a directory called `aa` and then the second build to `bb`. After this run `diffoscope` against them.

```
mkdir aa bb
PKG_PATH=/home/user/code/securedrop-client/dist/securedrop-client-0.0.6.tar.gz PKG_VERSION=0.0.6 make securedrop-client
mv ~/debbuild/packaging/securedrop-client_0.0.6_all.deb aa/

PKG_PATH=/home/user/code/securedrop-client/dist/securedrop-client-0.0.6.tar.gz PKG_VERSION=0.0.6 make securedrop-client
mv ~/debbuild/packaging/securedrop-client_0.0.6_all.deb bb/

$ diffoscope aa/securedrop-client_0.0.6_all.deb bb/securedrop-client_0.0.6_all.deb 
|###############################################################################################################################################################|  100%                             Time: 0:00:00 
```

### DIFF
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -45,6 +45,7 @@ fi
 # Disable use of pip cache during debhelper build actions.
 export DH_PIP_EXTRA_ARGS="--no-cache-dir --require-hashes"
 
+
 # Declare general packaging building workspace; subdirs will
 # be created within, to build specific packages.
 TOP_BUILDDIR="$HOME/debbuild/packaging"
@@ -72,6 +73,9 @@ cd "$TOP_BUILDDIR/$PKG_NAME/"
 $CUR_DIR/scripts/verify-hashes $CUR_DIR/sha256sums.txt ./requirements.txt
 
 echo "All hashes verified."
+
+# Adds reproducibility step
+export SOURCE_DATE_EPOCH=`dpkg-parsechangelog -STimestamp`
 
 # Build the package
 dpkg-buildpackage -us -uc
--- a/securedrop-client/debian/rules
+++ b/securedrop-client/debian/rules
@@ -6,4 +6,8 @@
 override_dh_virtualenv:
 	dh_virtualenv --python /usr/bin/python3.5 --setuptools -S --index-url https://dev-bin.ops.securedrop.org/simple
 
-
+override_dh_strip_nondeterminism:
+	find ./debian/ -type f -name '*.pyc' -delete
+	find ./debian/ -type f -name 'pip-selfcheck.json' -delete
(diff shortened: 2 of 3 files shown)


joshuathayer:
Hi! So I'm following the test instructions above, and while the commands complete successfully, `diffoscope` shows that the .deb's that result have quite a lot of differences: 
[diff.txt](https://github.com/freedomofpress/securedrop-debian-packaging/files/2687654/diff.txt)

Not sure if I've done something wrong, or if this is expected...

conorsch:
@joshuathayer Great catch. Can you run `lintian` on those debs and check for errors regarding the inclusion of `.pyc` files? Even with dh-virtualenv, we shouldn't be shipping those. See docs at https://dh-virtualenv.readthedocs.io/en/1.0/usage.html#cmdoption-extra-pip-arg in case we need to change.

joshuathayer:
@conorsch that seems like the right direction, yeah. the `diffoscope` diff is complaining about strings in `.pyc`'s, certainly. Those .pyc's all (?) seem to be in `__pycache__` directories, and `lintian` complains about that: `package-installs-python-pycache-dir`.

conorsch:
Thanks for confirming, @joshuathayer. Try a build with the recommended flag:

> `--extra-pip-arg “–no-compile”` to the override_dh_virtualenv section of the debian/rules file will disable the generation of pyc files.

and see if that gets you closer to full reproducibility.

joshuathayer:
Ah, never mind! I was testing this on master and getting those diffs. When I test on this branch, `diffoscope` reports no diffs! So, I can confirm that this branch seems to make reproducible builds. Super cool.

joshuathayer:
(I don't have write access to this repo, so I can't appove the PR!)

PR #36, opened 2019-04-24
Updates changelog for securedrop-proxy 0.1.3

kushaldas:
Now one can build `securedrop-proxy` `0.1.3` package.


securedrop-proxy/debian/changelog
```diff
@@ -1,3 +1,10 @@
+securedrop-proxy (0.1.3) UNRELEASED; urgency=medium
```
emkll: UNRELEASED here is a divergence from previous releases, though should not have any impact on apt. Might be worth revising this for the beta

kushaldas: Yup, will update it in the next push.

PR #40, opened 2019-05-09
update install file for the securedrop client

creviera:
## Description

When trying to run the client from the debian package install, Kushal and I ran into an error where it was trying to look for a couple alembic migration files that had been deleted. We ended making the changes in this PR in order to build the 0.0.7 release of the client to get passed this error.

## Test Plan

1. build debian package of the client with this changes to the install file
2. install
3. run and see no error about trying to find the alembic files that were removed in this PR

/cc @kushaldas @redshiftzero

### DIFF
--- a/securedrop-client/debian/changelog
+++ b/securedrop-client/debian/changelog
@@ -1,3 +1,9 @@
+securedrop-client (0.0.7) unstable; urgency=medium
+
+  * Update securedrop-sdk to 0.0.8 (#357)
+
+ -- Allie Crevier <allie@freedom.press>  Mon, 06 May 2019 16:01:16 -0400
+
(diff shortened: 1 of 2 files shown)


redshiftzero:
this looks good - there should also be an addition to the [changelog](https://github.com/freedomofpress/securedrop-debian-packaging/blob/master/securedrop-client/debian/changelog) file from releasing securedrop-client 0.0.7 - can you add that to this PR?

creviera:
Update
----

Opened follow-up issues for changelog process:

* https://github.com/freedomofpress/securedrop-debian-packaging/issues/41
* https://github.com/freedomofpress/securedrop-client/issues/367
* https://github.com/freedomofpress/securedrop-proxy/issues/31

PR #47, opened 2019-06-04
Fixes libreoffice grsec error

kushaldas:
The error was mentioned at
https://github.com/freedomofpress/securedrop-workstation/pull/263#issuecomment-498324182

Part of fixing https://github.com/freedomofpress/securedrop-workstation/issues/205

### DIFF
--- a/securedrop-workstation-svs-disp/paxctld.conf.svsdisp
+++ b/securedrop-workstation-svs-disp/paxctld.conf.svsdisp
@@ -107,11 +107,8 @@
 # openrc
 /lib/rc/bin/lsb2rcconf		E
 
+
 # libreoffice
-# Ubuntu doesn't seem to carry this patch:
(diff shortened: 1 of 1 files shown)


emkll:
Thanks @kushaldas !
In the initial review of https://github.com/freedomofpress/securedrop-workstation/pull/263, I had mis attributed the error I've observed to a missing pax flag, when in fact the issue pas the paxctld service was not running. Based on further review of that pr, which did not use the updated verison of this package, and [the comments](https://github.com/freedomofpress/securedrop-debian-packaging/pull/47/files#diff-ba94893640632cf046941cea20f283a8L111) in the paxctld.conf file, this pax flag is not strictly necessary. While we can still observe a line in the logs, LibreOffice (writer, calc, etc..) works fine.
I propose we close this PR and not merge at this time, we can re-open in the future if this poses further issues (if we do re-open, we should also bump the version/changelog of the package).

PR #43, opened 2019-05-18
Rework packaging scripts to remove pipenv

redshiftzero:
Closes #42. I've tried to be pretty verbose here in comments in the diff for maintenance sake. 

One can test these changes by going through the full packaging workflow with the following two pending PRs: 

- https://github.com/freedomofpress/securedrop-proxy/pull/33 (with commit f946fe38f4dc0b4b71cff3ea8d318fcca1599c09 dropped which fixes the CVEs in the dependencies in that repo (alternatively one can add those packages to FPF's PyPI after review))
- https://github.com/freedomofpress/securedrop-client/pull/372 (with commit 97cdd30117aa0e4dff331d586376a8da6e0eb9f4 dropped which fixes the CVEs in the dependencies in that repo (alternatively one can add those packages to FPF's PyPI after review))

Note: this PR is ready for review and will be taken off of draft mode once this and the above two PRs are approved - this is such that we can merge all three in a coordinated manner without breaking debian package builds

(CI is "failing" because I enabled circle CI here to add CI jobs in #44 but there's no config file in the master branch/this PR, so we should ignore until we merge #44)

PR #52, opened 2019-06-08
Update release manager steps to account for pipenv removal

redshiftzero:
I forgot to update the release manager steps in #43, here are the text changes to reduce confusion for release managers. diagram and gif still need updating or removing, which we should do in a separate PR


redshiftzero:
this is a tiny PR for you @kushaldas

PR #58, opened 2019-06-18
ci: increment version and build nightly securedrop-client deb package

redshiftzero:
Closes https://github.com/freedomofpress/securedrop-client/issues/407 (see rationale for versioning scheme [here](https://github.com/freedomofpress/securedrop-client/issues/407#issuecomment-503283197))

Contains securedrop-client side changes for https://github.com/freedomofpress/securedrop-debian-packaging/issues/50

Test build [here](https://circleci.com/gh/freedomofpress/securedrop-debian-packaging/57): observe the built package is in `/home/circleci/debbuild/packaging/securedrop-client_0.0.8-dev-20190618-205506_all.deb` 

This PR also contains some DRYing up so that the securedrop-proxy version of this change will be pretty minimal (to do once https://github.com/freedomofpress/securedrop-proxy/pull/37 is merged)

PR #59, opened 2019-06-20
securedrop-client.install: use wildcard for alembic migrations

### LINKED ISSUE #53: Use wildcard for migration paths in securedrop-client.install instead of hardcoding
We hardcode paths to the alembic migrations in `securedrop-client.install`.

This introduces a problem for:
1. A nightly job built of latest master branch in the client, which may or may not contain a new migration
2. A CI job built off the diff in a given PR in the securedrop-client repository, which again may or may not contain a new migration

According to [this doc](https://manpages.debian.org/testing/debhelper/dh_install.1.en.html#FILES), we can use wildcards so we should use them instead of hardcoding the individual alembic migration paths

redshiftzero:
Closes #53

Needed to address this because https://github.com/freedomofpress/securedrop-client/pull/427 is adding a new migration, and I don't want the nightly build to be added soon in #58 to start producing a non-working package when we merge that PR

For the lazy (this is how I tested this), the `build-securedrop-client` CI job here is handy, one can wait until the package is built in the `build-securedrop-client` job, SSH in to the circle CI box and:

0. `cd ~/debbuild/packaging`
1. `ar x <name_of_deb>`
2. `tar xvf data.tar.xz`
3. `ls usr/share/securedrop-client/alembic/versions/`
4. Confirm you see the migration, even though we didn't specify it manually

### DIFF
--- a/securedrop-client/debian/securedrop-client.install
+++ b/securedrop-client/debian/securedrop-client.install
@@ -2,6 +2,6 @@ files/alembic.ini usr/share/securedrop-client/
 alembic/env.py usr/share/securedrop-client/alembic/
 alembic/README usr/share/securedrop-client/alembic/
 alembic/script.py.mako usr/share/securedrop-client/alembic/
-alembic/versions/2f363b3d680e_init.py usr/share/securedrop-client/alembic/versions/
+alembic/versions/*.py usr/share/securedrop-client/alembic/versions/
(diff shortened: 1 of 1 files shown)


creviera:
I was able to confirm seeing the alembic migration file, even though it was not explicitly specified, so it appears to be working!

PR #64, opened 2019-06-28
ci: commit workstation debs to lfs repo after nightly build

### LINKED ISSUE #51: Modify nightly build logic to commit to LFS repository
- Part of build automation epic: #8
- Pending merge of #44
- Pending completion and integration of version bump logic (#50)
- Pending completion of infrastructure work to set up a new LFS repository for this purpose (tracked internally)

Once all these dependencies are met, we can update the nightly build job to commit to the LFS repository set up to store developer build artifacts (likely `securedrop-dev-packages-lfs` or similar). Completion of this task requires CircleCI administrator privileges to configure the credentials.

redshiftzero:
Closes #51

This PR makes the nightly workflow first build securedrop-proxy, then commit to master in the dev packages LFS repo, then do the same for securedrop-client.

One can observe the bot account commit packages here: https://github.com/freedomofpress/securedrop-dev-packages-lfs/commits/master

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,39 @@ common-steps:
         make $PKG_NAME
         ls ~/debbuild/packaging/*.deb
 
+  - &installgitlfs
+    run:
+      name: Install Git LFS.
+      command: |
+        export GIT_LFS_VERSION=2.7.2
+        export GIT_LFS_CHECKSUM=89f5aa2c29800bbb71f5d4550edd69c5f83e3ee9e30f770446436dd7f4ef1d4c
+        wget https://github.com/git-lfs/git-lfs/releases/download/v$GIT_LFS_VERSION/git-lfs-linux-amd64-v$GIT_LFS_VERSION.tar.gz
+        sha256sum git-lfs-linux-amd64-v$GIT_LFS_VERSION.tar.gz | grep $GIT_LFS_CHECKSUM
+        tar xzf git-lfs-linux-amd64-v$GIT_LFS_VERSION.tar.gz
+        sudo mv git-lfs /usr/local/bin/git-lfs
+        git lfs install
+
+  - &addsshkeys
+    add_ssh_keys:
+      fingerprints:
+        - "e5:b5:6e:d0:4e:ce:52:40:33:30:5e:6f:c5:73:38:20"
+
+  - &commitworkstationdebs
+    run:
+      name: Commit workstation debs for deployment to apt-test-qubes.freedom.press
+      command: |
+        git clone git@github.com:freedomofpress/securedrop-dev-packages-lfs.git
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -81,6 +81,42 @@ common-steps:
...
+        git config user.name "sdcibot"
+
+        # BUILD_BRANCH is master for automated builds
+        BUILD_BRANCH=master git checkout $BUILD_BRANCH
+
+        # Copy built debian packages to the relevant workstation repo and git push.
+        cp ~/debbuild/packaging/*.deb ./workstation/stretch/
+        git add workstation/stretch/*.deb
+        git commit -m "Automated SecureDrop workstation build"
+        git push origin $BUILD_BRANCH
```
conorsch: There are two uses of `$BUILD_BRANCH` in the commit logic, and in neither case is a value defined when the variable is referenced. Exporting the value on its own line would resolve. As illustration of the problem:

```
$ foo=bar echo $foo
$
$ echo $foo
$
```

In neither case is an error thrown, but it's a silent failure. If we stuck this logic in a script with `set -u` then we'd catch undefined errors in the future. Tangentially, check out this snarky gem from `man git-checkout`:

>  You could omit \<branch\>, in which case the command degenerates to "check out the current branch", which is a glorified no-op with a rather expensive side-effects to show only the tracking information, if exists, for the current branch.

The fact that `git push origin` is run instead of `git push origin master` isn't truly a problem for the CI use case, but still, best to be keep it explicit with the master branch as refspec.

redshiftzero: ahh yeah - ok I've removed this variable entirely because it's unnecessary

.circleci/config.yml
```diff
@@ -81,6 +81,39 @@ common-steps:
...
+        - "e5:b5:6e:d0:4e:ce:52:40:33:30:5e:6f:c5:73:38:20"
+
+  - &commitworkstationdebs
+    run:
+      name: Commit workstation debs for deployment to apt-test-qubes.freedom.press
+      command: |
+        git clone git@github.com:freedomofpress/securedrop-dev-packages-lfs.git
+        cd securedrop-dev-packages-lfs
+
+        git config user.email "securedrop@freedom.press"
```
conorsch: The email address will need to change if/when we implement commit signing. Fine for now!

.circleci/config.yml
```diff
@@ -148,5 +187,7 @@ workflows:
               only:
                 - master
     jobs:
-      - build-nightly-securedrop-client
       - build-nightly-securedrop-proxy
+      - build-nightly-securedrop-client:
+          requires:
+             - build-nightly-securedrop-proxy
```
conorsch: Great use of dependency on the proxy package. :+1:

Issue #63, opened 2019-06-26
Build securedrop core release candidate packages in CI here


redshiftzero:
We want the following process to automate some RM tasks in SecureDrop core:

1. Release manager pushes an rc tag in SecureDrop core
2. CI builds all rc packages
3. CI commit packages to lfs repo for automatic deployment to apt-test.freedom.press

The issue is that we can't put the git commit bot credentials in Circle CI in the core SecureDrop repository. We can here though since forks don't get access to Circle CI creds (indeed forked builds are off entirely). 

A proposed workaround is to add a step, which is that after a release manager pushes an rc tag in securedrop core, they then push a tag here to trigger a build called `securedrop-core-$TAG` where `$TAG` is the tag to be built. We would use CircleCI's tag filtering logic to check that tag out and build, e.g. something like in the CI config:

```
      - build-securedrop-core-release-candidate:
          filters:
            tags:
              only:
                - /securedrop-core-*rc*/
```

If you see a better solution here to trigger builds on tag that does not involve giving the SecureDrop core repo credentials, please comment.

cfm:
https://github.com/freedomofpress/securedrop-debian-packaging/issues/63#issue-461147738:

> If you see a better solution here to trigger builds on tag that does not involve giving the SecureDrop core repo credentials, please comment.

One possibility would be to have `securedrop`'s `-*rc*` workflow trigger a run of `build-securedrop-core` here [via the CircleCI API](https://circleci.com/docs/2.0/triggers/#trigger-a-job-using-curl-and-your-api-token).  That would eliminate both the presence of `securedrop` code in this repository's tags and the need for the Release Manager to kick off this process manually, requiring just a `$CIRCLE_API_USER_TOKEN` in `securedrop`.

I'd be happy to pick this up as a CI priority in an upcoming sprint.

PR #65, opened 2019-06-28
[wip] ci: build securedrop rc packages

redshiftzero:
After #64 (automatic nightly build/push of client/proxy), then we can do #63 (automatic build/push of core release candidates), this is work towards that, putting up wip in case anyone has strong feelings on this

Remaining tasks:
 - Reuse steps in #64 and commit only _untracked files_ (to handle the fact that the deploy logic (correctly) doesn't republish packages with unchanged versions but changed checksums) to `core/xenial/` instead of `workstation/stretch/`
 - Add [tag filter](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag) to trigger `build-securedrop-core` only on `securedrop-core/*` tag push (off for now while testing, the behavior in circle is that it will add the tag to `$CIRCLE_TAG` which you'll see is being used in the below diff)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,15 @@ common-steps:
         echo $PKG_NAME > ~/packaging/sd_package_name
         echo 'export PKG_NAME=$(cat ~/packaging/sd_package_name)' >> $BASH_ENV
 
+  - &clone_and_build_securedrop_core
+    run:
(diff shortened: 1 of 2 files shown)


legoktm:
I'm going to close this because it's pretty out of date now and while we do want to move in this direction, it would be done in the securedrop server repo, not this one.

PR #69, opened 2019-07-22
Adds securedrop-export build logic

zenmonkeykstop:
Adds build logic for the securedrop-export debian package, which contains submission export scripts used by the Qubes Workstation.

To build:

0. Set up build environment as per this repo's README
1. Create a source tarball using this branch of securedrop-export: https://github.com/freedomofpress/securedrop-export/tree/271-package-send-to-usb
2. set PKG_VERSION: `export PKG_VERSION=0.1.1`
3. run make: `make securedrop-export`


securedrop-export/debian/control
```diff
@@ -0,0 +1,17 @@
...
+Maintainer: SecureDrop Team <securedrop@freedom.press>
+Build-Depends: debhelper (>= 9), dh-python, python3-all, python3-setuptools, dh-virtualenv
+Standards-Version: 3.9.8
+Homepage: https://github.com/freedomofpress/securedrop-export
+Vcs-Git: https://github.com/freedomofpress/securedrop-export.git
+X-Python-3-Version: >= 3.5
+
+Package: securedrop-export
+Architecture: all
+Depends: ${python3:Depends}, ${misc:Depends}
```
emkll: Would it make sense to add the dependencies that are strictly required for this script to work (eg, printer drivers, xpp, cups)?
https://github.com/freedomofpress/securedrop-workstation/blob/503fd6f0930a24ddc92d9cfae4fa502275b3362f/dom0/sd-export-files.sls#L15:L25

zenmonkeykstop: Sure, although would they be better off in a metapackage? (might be getting into gilding the lily a bit at that stage tho.)

securedrop-export/debian/postinst
```diff
@@ -0,0 +1,43 @@
...
+#          <failed-install-package> <version> `removing'
+#          <conflicting-package> <version>
+# for details, see https://www.debian.org/doc/debian-policy/ or
+# the debian-policy package
+
+
+case "$1" in
+    configure)
+    
+    update-desktop-database /usr/share/applications
```
emkll: I believe the mime database also needs to be updated as well, see https://github.com/freedomofpress/securedrop-workstation/blob/503fd6f0930a24ddc92d9cfae4fa502275b3362f/dom0/sd-export-files.sls#L67

PR #71, opened 2019-07-24
Add nightly builds for securedrop-export package

emkll:
Adds nightly build and commits to LFS  deb packages for securedrop-export (see https://github.com/freedomofpress/securedrop-debian-packaging/pull/64 and https://github.com/freedomofpress/securedrop-debian-packaging/pull/61)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,17 @@ common-steps:
         echo $PKG_NAME > ~/packaging/sd_package_name
         echo 'export PKG_NAME=$(cat ~/packaging/sd_package_name)' >> $BASH_ENV
 
+  - &clonesecuredropexport
+    run:
+      name: Clone the repository to be packaged
+      command: |
+        mkdir ~/packaging && cd ~/packaging
+        git clone https://github.com/freedomofpress/securedrop-export.git
+        export PKG_NAME="securedrop-export"
+        # Enable access to this env car in subsequent run steps
+        echo $PKG_NAME > ~/packaging/sd_package_name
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -63,6 +63,19 @@ common-steps:
...
         echo 'export PKG_NAME=$(cat ~/packaging/sd_package_name)' >> $BASH_ENV
 
+  - &clonesecuredropexport
+    run:
+      name: Clone the repository to be packaged
+      command: |
+        mkdir ~/packaging && cd ~/packaging
+        git clone https://github.com/freedomofpress/securedrop-export.git
+        # Checkout WIP branch
+        cd securedrop-export && git checkout add-makefile-ci
```
zenmonkeykstop: `add-makefile-ci` was merged into master (CI is failing due to a subsequent upstream module deletion), so this is probably unnecessary

emkll: good catch, fixed (I made a mistake while rebasing earlier)

.circleci/config.yml
```diff
@@ -191,3 +231,4 @@ workflows:
       - build-nightly-securedrop-client:
           requires:
              - build-nightly-securedrop-proxy
+      - build-nightly-securedrop-export
```
redshiftzero: flagging: this should be changed before the next nightly build as it will likely fail - this new nightly job should require `build-nightly-securedrop-proxy` and `build-nightly-securedrop-client` so that these jobs run in sequence, not concurrently, as they will run in about the same amount of time and commit directly to master in the LFS repo

PR #72, opened 2019-07-24
Adds sd-svs mimetype handlers to securedrop-client pkg

conorsch:
The files referenced here were added to the manifest file in the
"securedrop-client" repository, ensuring they are included in the
tarball used for creating a Debian package.

See corresponding client changes in https://github.com/freedomofpress/securedrop-client/pull/490

### Testing

0. Set up build environment as per this repo's README
1. Create a source tarball using this branch of securedrop-client: https://github.com/freedomofpress/securedrop-client/pull/490
2. set PKG_VERSION: `export PKG_VERSION=0.0.8`
3. run make: `make securedrop-client`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ build-wheels: fetch-wheels ## Builds the wheels and adds them to the localwheels
 
 .PHONY: clean
 clean: ## Removes all non-version controlled packaging artifacts
+	rm -rf localwheels/*
 
 .PHONY: help
 help: ## Prints this message and exits
--- a/securedrop-client/debian/changelog
+++ b/securedrop-client/debian/changelog
@@ -1,3 +1,16 @@
+securedrop-client (0.0.9) unstable; urgency=medium
+
+  * Use Montserrat and Source Sans Pro (#493)
+  * Same BG for scrollarea, conversation view, selected source (#494,#496)
+  * Supports opening submissions in DispVMs from Qubes dev env (#490)
+  * Center all popup windows (#487)
+  * Use priority queue for job processing (#486)
+  * Use SecureQLabel (#485)
+  * Extract and display original document filenames (#452)
+  * Save in-progress replies via persisting SourceConversationWrapper (#431)
+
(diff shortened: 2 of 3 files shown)


emkll:
rebased on latest `master`

emkll:
Visual diff looks good to me, and local testing is successful. However CI is legitimately failing, on `build-securedrop-client`.

The issue is that this failing CI job checks out the latest release tag (in this case 0.0.8). Because files we added to the tarball and referenced explicitly in the .install file, these changes need to happen in both repos at the same time to keep CI green.
I see 3 ways of resolving this:
* Modify the build-debian-packages CI workflow to use check out `master` instead of the latest release (tag) prior to building tarball [1] .
* Cut a new securedrop-client release whenever there are changes in the securedrop-client code that require changes in the packaging logic.
* Use wildcards for files but that sounds like a bad idea.

The first option sounds preferable to me, but we should probably discuss this in standup or engineering meeting.

[1] https://circleci.com/gh/freedomofpress/securedrop-debian-packaging/276?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

redshiftzero:
I think we want to use this repository to always be able to:

* build latest tagged release of each project (without stepping through the git history to see which version of the build logic we should be using) - tested via these three CI jobs on PR merge
* build latest `master` of each project using latest `master` of the relevant project (for nightlies)

If you agree, @emkll, that means that we should do 2: cut a release of client when there are changes that require corresponding packaging changes (we could do something like tag the securedrop-debian-packaging repo and be clear about which versions build on which version of the debian packaging logic but since we care about the latest version I think it's easier to just tag client when we need to). What do you think?

emkll:
Thanks @redshiftzero , that definitely makes sense. Client 0.0.9 release was merged in  https://github.com/freedomofpress/securedrop-client/pull/499 , CI is now passing :tada: . I will push an updated changelog for the debian package to this branch for review prior to merge.

redshiftzero:
approving commit ac9c0ac, lgtm

PR #80, opened 2019-10-10
doc: make install-deps needs to be run in the template

redshiftzero:


README.md
```diff
@@ -23,6 +23,9 @@ make install-deps
 make fetch-wheels
 ```
 
+**Note:** either run `make install-deps` each time you start your debian packaging AppVM, or make
```
creviera: ah, yes, yes. this.

PR #85, opened 2019-11-04
Updates the script to add multiple wheel hashes

kushaldas:
Now we can have multiple wheels hashes into the build-requirement
script, for python3.5 and python3.7

## How to test?

On a Buster VM/system

- [ ] `PKG_DIR=/home/user/code/securedrop-client make requirements`
- [ ] `PKG_DIR=/home/user/code/securedrop-client make requirements`

This should create a build-requirements.txt with multiple hashes for sqlalchemy and MarkupSafe wheels.

### DIFF
--- a/scripts/update-requirements
+++ b/scripts/update-requirements
@@ -76,14 +76,15 @@ def add_sha256sums(path: str, requirements_lines: str) -> None:
     missing_wheels = []
 
     # For each dependency in the requirements file
-    for line in requirements_lines:
-        package_name_and_version = line.strip().split()[0]
+    for mainline in requirements_lines:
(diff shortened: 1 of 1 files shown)


redshiftzero:
this should be reviewed after #86, correct? (since the sha256sums.txt on this branch only has one wheel per dependency)

redshiftzero:
Due to the `make requirements` checks that we have in CI (that ensure that the build requirements are up to date given what is in `requirements.txt`), when we changed `make requirements` to produce different build requirements - i.e. since `build-requirements.txt` should have wheels for both 3.5 and 3.7 - we _also_ need to update the `build-requirements.txt` in repos with this check or CI build jobs will fail

PR #97, opened 2019-11-21
add build job for python3.7-buster

creviera:
## Description

Adds a `securedrop-proxy` build job for Python 3.7 on Buster.

To test, build the `securedrop-proxy` debian package in a Buster VM and run through test plan to make sure it is functioning as normal without regression.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ jobs:
       - checkout
       - run:
           name: install test requirements and run tests
-          command: |   
+          command: |
             virtualenv .venv
             source .venv/bin/activate
             pip install -r test-requirements.txt
@@ -194,7 +194,7 @@ jobs:
 
   build-securedrop-proxy-buster:
     docker:
-      - image: circleci/python3.7-buster
+      - image: circleci/python:3.7-buster
     steps:
       - checkout
       - *installdeps
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,9 +180,21 @@ jobs:
       - *addsshkeys
       - *commitworkstationdebs
 
-  build-securedrop-proxy:
+  build-securedrop-proxy-stretch:
     docker:
       - image: circleci/python:3.5-stretch
+    steps:
+      - checkout
+      - *installdeps
+      - *fetchwheels
+      - *clonesecuredropproxy
+      - *getlatestreleasedversion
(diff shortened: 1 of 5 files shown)


securedrop-proxy/debian/securedrop-proxy.triggers
```diff
@@ -1,7 +1,7 @@
 # Register interest in Python interpreter changes (Python 2 for now); and
 # don't make the Python package dependent on the virtualenv package
 # processing (noawait)
-interest-noawait /usr/bin/python3.5
```
redshiftzero: ah so only the changes in the `securedrop-proxy/` directory will be needed here

.circleci/config.yml
```diff
@@ -192,6 +192,18 @@ jobs:
       - *makesourcetarball
       - *builddebianpackage
 
+  build-securedrop-proxy-buster:
+    docker:
+      - image: circleci/python3.7-buster
```
redshiftzero: ahh `circleci/python3.7-buster` -> `circleci/python:3.7-buster`

redshiftzero: (that's why CI was failing, this was confusing me)

creviera: thanks! :star2: for you

securedrop-proxy/debian/securedrop-proxy.triggers
creviera: whoops went a little overboard with find-and-replace

securedrop-proxy/debian/changelog-buster
```diff
@@ -0,0 +1,5 @@
+securedrop-proxy (0.1.5) unstable; urgency=medium
```
redshiftzero: ah so this version `0.1.5` should end in `+buster` i.e. `0.1.5+buster` if you are running on the latest version of `master` in this repository - can you double check that you generated the changelog on the latest `master`?

creviera:
@redshiftzero I'm still in the process of building the package on buster, but I'm seeing a makefile error: `Error 1`

debugging now, but wanted to update here in case you are able to test and there's something wrong with my new deb-packaging-buster vm

creviera:
`make fetch-wheels` fixed my issue, i just need to add entries to both the stretch and (new!) buster changelogs and then this will be ready for review

PR #104, opened 2019-12-06
Adds initial packaging for securedrop-log

kushaldas:
Adds Buster package for the securedrop-log package.

## How to test?

- [ ] In the securedrop-log directory, run `python3 setup.py sdist`
- [ ] `PKG_VERSION=0.0.1 PKG_PATH=/home/user/project/securedrop-log/dist/securedrop-log-0.0.1.tar.gz make securedrop-log`  Remember to change the path based on your actual code path.


securedrop-log/debian/securedrop-log.install
```diff
@@ -0,0 +1 @@
+securedrop.Log etc/qubes-rpc/securedrop.Log
```
redshiftzero: so `securedrop.Log` isn't yet in `securedrop-log` (though it's described in the README how to create) - we should add a file over there called `securedrop.Log` containing just `/usr/sbin/securedrop-log` and then this should work

kushaldas: Wait, I thought I pushed that in a PR last night. Let me check

kushaldas: Yup, I forgot to create the PR (it was after midnight I guess) https://github.com/freedomofpress/securedrop-log/pull/3

securedrop-log/debian/control
```diff
@@ -0,0 +1,16 @@
...
+Maintainer: SecureDrop Team <securedrop@freedom.press>
+Build-Depends: debhelper (>= 9), dh-python, python3-all, python3-setuptools, dh-virtualenv
+Standards-Version: 3.9.8
+Homepage: https://github.com/freedomofpress/securedrop-log
+Vcs-Git: https://github.com/freedomofpress/securedrop-log.git
+X-Python-3-Version: >= 3.7
+
+Package: securedrop-log
+Architecture: all
+Depends: ${python3:Depends}, ${misc:Depends},  python3-dev
```
redshiftzero: why do we need `python3-dev`?

securedrop-log/debian/securedrop-log.install
redshiftzero: I think this will create a _folder_, i.e. `/etc/qubes-rpc/securedrop.Log/securedrop.Log`, instead what about:

```suggestion
securedrop.Log etc/qubes-rpc/
```

securedrop-log/debian/control
kushaldas: typo

securedrop-log/debian/securedrop-log.install
kushaldas: fixing this.

kushaldas:
@redshiftzero Those two new commits look good to me. Approved.

PR #105, opened 2019-12-10
AppArmor support and profile for securedrop-client

emkll:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/234 :

* Bump kernel image to support AppArmor (see changes in https://github.com/freedomofpress/ansible-role-grsecurity-build/pull/54)
* Logic to copy, squash and enforce AppArmor profile for securedrop-client

### Test plan
subset of https://github.com/freedomofpress/securedrop-workstation/pull/364 , required a vm with apparmor support:

- [ ] install `apparmor-utils` on sd-svs or sd-svs-buster-template
- [ ] build securedrop-client using this branch and https://github.com/freedomofpress/securedrop-client/pull/673
- [ ] Installing the deb results in an apparmor profile enforced the host (`sudo aa-status`)

### DIFF
--- a/securedrop-client/debian/postinst
+++ b/securedrop-client/debian/postinst
@@ -22,6 +22,8 @@ case "$1" in
     configure)
 
     update-desktop-database /usr/share/applications
+    # enforce the apparmor profile for the client
+    aa-enforce /etc/apparmor.d/usr.bin.securedrop-client
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
--- a/securedrop-client/debian/control
+++ b/securedrop-client/debian/control
@@ -9,5 +9,5 @@ X-Python3-Version: >= 3.5
 
 Package: securedrop-client
 Architecture: all
-Depends: ${python3:Depends},${misc:Depends}, python3-pyqt5, python3-pyqt5.qtsvg
+Depends: ${python3:Depends},${misc:Depends}, python3-pyqt5, python3-pyqt5.qtsvg, apparmor-utils
 Description: securedrop client for qubes workstation
--- a/securedrop-client/debian/securedrop-client.install
+++ b/securedrop-client/debian/securedrop-client.install
@@ -8,3 +8,4 @@ files/open-in-dvm.desktop usr/share/applications/
 files/sd-svs-qubes-gpg-domain.sh etc/profile.d/
 files/securedrop-client usr/bin/
 files/securedrop-client.desktop usr/share/applications/
+files/usr.bin.securedrop-client /etc/apparmor.d/


securedrop-client/usr.bin.securedrop-client
```diff
@@ -0,0 +1,75 @@
+# Last Modified: Tue Dec 10 11:57:59 2019
+#include <tunables/global>
+
+/usr/bin/securedrop-client {
+  #include <abstractions/base>
+  #include <abstractions/fonts>
+  #include <abstractions/python>
+  #include <abstractions/user-tmp>
+
+  deny /usr/bin/sudo x,
```
emkll: This explicit deny is not strictly required

securedrop-client/debian/postinst
```diff
@@ -22,6 +22,17 @@ case "$1" in
     configure)
 
     update-desktop-database /usr/share/applications
+    # If the profile was disabled enabled it.
+    if [ -e "/etc/apparmor.d/disable/usr.bin.securedrop-client" ]; then
+        rm /etc/apparmor.d/disable/usr.bin.securedrop-client
+    fi
+
+    # if the deb package is providing a new version of the apparmor profile, squash the old one
```
rmol: Why is this explicit handling necessary?

redshiftzero:
based on visual review of the diff, this looks great

redshiftzero:
if a developer makes a change in securedrop-client, they will not notice breakage introduced if the apparmor profile is not updated in this repository - it will only be detected when we try to run the nightly build. two ideas from discussing this:
* add a note to the PR template in securedrop-client indicating that a change may require updating the apparmor profile, the reviewer should ideally test in the "production" environment (for small changes it may be skipped, as we do with merging small webapp changes in securedrop core)
* @emkll suggested adding a functional test that installs the package and starts the client (in securedrop-client CI)

securedrop-client/debian/postinst
emkll: Just tested locally, running aa-enforce will reliably remove the symlink from `/etc/apparmor.d/disabled/` . I will remove these lines.

emkll:
Rebased and resolved conflicts from latest `master`, addressed https://github.com/freedomofpress/securedrop-debian-packaging/pull/105#discussion_r357215394.

Opened https://github.com/freedomofpress/securedrop-client/issues/665 to follow up on the comments raised in https://github.com/freedomofpress/securedrop-debian-packaging/pull/105#issuecomment-566137365

I don't think we should block merge and address the above separately, as part of the current sprint, as this PR is blocking https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/30 . However, happy to split up the PR if necessary.

kushaldas:
@emkll Quick question: Why is the actual apparmor rules file is a part of this repo but not the  source repo of `securedrop-client`?

This current implementation is following the same pattern of `securedrop-app-code` packaging, where the files were kept in the same repo. 

As we separated the source and debian packaging via introducing this repository, I feel we should follow that, and move the  `usr.bin.securedrop-client ` to the `securedrop-client` repo instead. The idea was that developers will do a source tar ball release, and then the packager (another persona), will just work on the packaging, without introducing any addition to the package.

@redshiftzero @conorsch thoughts?

emkll:
@kushaldas good point. Initially, I had thought of putting the apparmor profile in the securedrop-client repository  initially. Since developers can't really use the profile (due to the paths being different in dev and prod), there was no benefit to having the AppArmor profile in that location.

You make a good point that the profile is more an app-level configuration than a packaging level configuration: If we were to make a change in prod for an apparmor profile change only, we would need to bump the securedrop-client version. There would be no advantage in having the apparmor profile in here.

emkll:
Added apparmor profile in https://github.com/freedomofpress/securedrop-client/pull/673, which will need to be merged first. We will then need to release a new Client version to ensure the apparmor profile is in place. Afterwards, we should re-run CI on this PR for it to pass.

emkll:
This PR has now been scoped to securedrop-client related apparmor changes, kernel changes are in #116 . The order of operations can be found in https://github.com/freedomofpress/securedrop-workstation/pull/364

emkll:
https://github.com/freedomofpress/securedrop-client/pull/673 has been merged and CI is passing here, this is now ready for final review.

rmol:
:+1: Installed the package in `sd-svs-buster-template`, `apparmor-utils` was automatically installed, `/etc/apparmor.d/usr.bin.securedrop-client` is present and enforced, and the client works normally in `sd-svs`.

PR #112, opened 2019-12-12
Improve dh-virtualenv for securedrop-client

### LINKED ISSUE #111: Make dh_virtualenv more repeatable/reliable
The way we're invoking `dh_virtualenv` can result in Debian packages that are broken when deployed because of variation in the build environment. 

For example, having the Debian package `rhythmbox-plugins` installed in the builder VM pulls in the `python3-mako` package. When `dh_virtualenv` is passed `--system-site-packages` without `--ignore-installed` in the construction of the `securedrop-client` Debian package, `dh_virtualenv` decides the `mako` requirement is satisfied and doesn't install it in the virtualenv. When the resulting `.deb` is installed in `sd-svs`, where the `python3-mako` package is not installed, the SecureDrop client breaks because `mako` is unavailable.

The solution is to add some options to `dh_virtualenv` in the various `debian/rules` files:
```
--extra-pip-arg "--ignore-installed" \
--extra-pip-arg "--no-deps" \
--extra-pip-arg "--no-cache-dir" \
```

The importance of these for repeatable builds is explained in the [dh_virtualenv docs](https://dh-virtualenv.readthedocs.io/en/latest/usage.html#pbuilder-and-dh-virtualenv).

rmol:
Add extra pip options "--ignore-installed", "--no-deps", and "--no-cache-dir" to the dh-virtualenv override in rules files for securedrop-client, securedrop-export, and securedrop-proxy. This prevents variation in the build environment from changing the set of requirements included in the virtualenv embedded in the package.

Fixes #111.

### DIFF
--- a/securedrop-client/debian/rules
+++ b/securedrop-client/debian/rules
@@ -4,7 +4,15 @@
 	dh $@ --with python-virtualenv
 
 override_dh_virtualenv:
-	dh_virtualenv --python /usr/bin/python3 --setuptools -S --index-url https://dev-bin.ops.securedrop.org/simple --requirements build-requirements.txt
+	dh_virtualenv \
+		--python /usr/bin/python3 \
+		--setuptools \
+		--use-system-packages \
+		--index-url https://dev-bin.ops.securedrop.org/simple \
+		--extra-pip-arg "--ignore-installed" \
+		--extra-pip-arg "--no-deps" \
+		--extra-pip-arg "--no-cache-dir" \
+		--requirements build-requirements.txt
 
 override_dh_strip_nondeterminism:
 	find ./debian/ -type f -name '*.pyc' -delete
--- a/securedrop-export/debian/rules
+++ b/securedrop-export/debian/rules
@@ -1,11 +1,20 @@
 #!/usr/bin/make -f
 
 %:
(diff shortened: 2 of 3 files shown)


securedrop-client/debian/rules
```diff
@@ -4,7 +4,15 @@
 	dh $@ --with python-virtualenv
 
 override_dh_virtualenv:
-	dh_virtualenv --python /usr/bin/python3 --setuptools -S --index-url https://dev-bin.ops.securedrop.org/simple --requirements build-requirements.txt
+	dh_virtualenv \
+		--python /usr/bin/python3 \
+		--setuptools \
+		--use-system-packages \
+		--index-url https://dev-bin.ops.securedrop.org/simple \
+		--extra-pip-arg "--ignore-installed" \
```
redshiftzero: very good catch @rmol, we should def add these three extra pip args, `--ignore-installed` in particular

@kushaldas what do you think of this change?

kushaldas:
These are the corresponding docs for `pip` extra args:

- <https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-deps>
- <https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-i>
- <https://pip.pypa.io/en/stable/reference/pip_install/#wheel-cache>

All of these look good to me, and testing one package, we should apply these to all the other
packages.

@rmol @redshiftzero

redshiftzero:
hey @rmol when you get a chance feel free to update this PR to make this same change in all python-based packages in this repo and take this out of draft status

PR #121, opened 2020-01-07
Update copyright and maintainer information

### LINKED ISSUE #108: review and clean up inconsistent maintainer entries, changelogs and copyright files
We have some clean up to do from the alpha days prior to the production workstation release: 

* All packages should have maintainer: `SecureDrop Team <securedrop@freedom.press>`

* copyright files should be accurate (e.g. https://github.com/freedomofpress/securedrop-debian-packaging/blob/master/securedrop-client/debian/copyright#L2 corresponds to the wrong dependencies and the wrong package)

* changelogs should not have any invalid entries in them (e.g. "Tues" is considered an invalid day and each time we do a build we get a warning: https://github.com/freedomofpress/securedrop-debian-packaging/blob/master/securedrop-proxy/debian/changelog-stretch#L12)

This also needs to be done for any RPMs. Please open a new ticket if you want to tackle the debian packaging parts without the RPM parts.

emkll:
Closes #108 

- Standardizes maintainer information
- updated control files to ensure proper upsteam repos are used
- fixes typos in historical changelogs
- sets default `DEBFULLNAME` and `DEBEMAIL` to `SecureDrop Team <securedrop@freedom.press>`
- adds LICENCE file (copied from securedrop-client repository)

### DIFF
--- a/securedrop-log/debian/copyright
+++ b/securedrop-log/debian/copyright
@@ -2,11 +2,11 @@ Upstream-Name: securedrop-log
 Source: https://github.com/freedomofpress/securedrop-log
 
 Files: *
-Copyright: 2019 Freedom of the Press Foundation
+Copyright: 2020 Freedom of the Press Foundation
 License: GPL-3.0+
 
 Files: debian/*
-Copyright: 2019 Freedom of the Press Foundation <securedrop@freedom.press>
+Copyright: 2020 Freedom of the Press Foundation <securedrop@freedom.press>
 License: GPL-3.0+
 
 License: GPL-3.0+
--- a/securedrop-proxy/debian/copyright
+++ b/securedrop-proxy/debian/copyright
@@ -3,46 +3,5 @@ Upstream-Name: securedrop-proxy
 Source: https://github.com/freedomofpress/securedrop-proxy
 
 Files: *
-Copyright: 2018 Freedom of the Press Foundation <securedrop@freedom.press>
+Copyright: 2020 Freedom of the Press Foundation <securedrop@freedom.press>
 License: GPLv3+
-
-Files: orderedmultidict*
-Copyright: Please check the project's home page.
-License: Unlicense
-
-Files: idna*
-Copyright: Please check the project's home page.
-License: BSD-like
-
-Files: certifi*
-Copyright: Please check the project's home page.
-License: MPL-2.0
-
-Files: requests*
-Copyright: Please check the project's home page.
-License: Apache 2.0
-
-Files: urllib3*
-Copyright: Please check the project's home page.
-License: MIT
-
-Files: werkzeug*
-Copyright: Please check the project's home page.
-License: BSD
-
-Files: pyyaml*
-Copyright: Please check the project's home page.
-License: MIT
-
-Files: chardet*
-Copyright: Please check the project's home page.
-License: LGPL
-
-Files: six*
-Copyright: Please check the project's home page.
-License: MIT
-
-Files: furl*
-Copyright: Please check the project's home page.
-License: Unlicense
-
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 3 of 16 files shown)


securedrop-export/debian/copyright
```diff
@@ -2,26 +2,5 @@ Upstream-Name: securedrop-export
...
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU General Public License for more details.
- .
- You should have received a copy of the GNU General Public License
- along with this program. If not, see <https://www.gnu.org/licenses/>.
- .
- On Debian systems, the complete text of the GNU General
- Public License version 3 can be found in "/usr/share/common-licenses/GPL-3".
+Copyright: 2020 Freedom of the Press Foundation
+License: AGPL-3.0+
```
redshiftzero: woops noticed this was not AGPL 3: https://github.com/freedomofpress/securedrop-export/pull/45

securedrop-log/debian/copyright
```diff
@@ -2,26 +2,5 @@ Upstream-Name: securedrop-log
...
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU General Public License for more details.
- .
- You should have received a copy of the GNU General Public License
- along with this program. If not, see <https://www.gnu.org/licenses/>.
- .
- On Debian systems, the complete text of the GNU General
- Public License version 3 can be found in "/usr/share/common-licenses/GPL-3".
+Copyright: 2020 Freedom of the Press Foundation <securedrop@freedom.press>
+License: AGPL-3.0+
```
redshiftzero: same situation as securedrop-export, we need to update this license

securedrop-proxy/debian/copyright
```diff
@@ -3,46 +3,5 @@ Upstream-Name: securedrop-proxy
...
-Files: six*
-Copyright: Please check the project's home page.
-License: MIT
-
-Files: furl*
-Copyright: Please check the project's home page.
-License: Unlicense
-
+Copyright: 2020 Freedom of the Press Foundation <securedrop@freedom.press>
+License: AGPLv3+
```
redshiftzero: license needs updating to AGPLv3 to make this change

securedrop-proxy/debian/copyright
```diff
@@ -3,46 +3,5 @@ Upstream-Name: securedrop-proxy
 Source: https://github.com/freedomofpress/securedrop-proxy
 
 Files: *
-Copyright: 2018 Freedom of the Press Foundation <securedrop@freedom.press>
-License: GPLv3+
-
-Files: orderedmultidict*
```
redshiftzero: looking at the examples in https://www.debian.org/doc/packaging-manuals/copyright-format/1.0 it does appear that we should be pointing to the contact information and licenses of upstream dependencies. This is a lot of manual editing and we'd need to maintain it as we update dependencies. We could do something during the debian package build script, inspect `setup.py` in each dependency and add the contents of the `license`, `author`, and `author_email` to `copyright` since we have all the tarballs downloaded. Let's do that (or something like it in case someone else has a better idea) in a followup (not for the beta)?

securedrop-log/debian/copyright
redshiftzero: https://github.com/freedomofpress/securedrop-log/pull/9

securedrop-proxy/debian/copyright
redshiftzero: https://github.com/freedomofpress/securedrop-proxy/pull/56

kushaldas:
Q: Why do all the applications/libraries have to be AGPL? My understanding is that we use that license for web applications, SecureDrop core is one such application. The standard libraries and tools should be on GPLv3+ as I think.

securedrop-proxy/debian/copyright
kushaldas: This is a good suggestion, but, we have to hope that all of our dependencies have license information in their `setup.py` files. A few years ago, [the story was different](https://anweshadas.in/a-walk-of-licenses-in-pypi/).

redshiftzero:
hmm that's a good point (TIL, thanks for the link!), @emkll seems like what we should do is modify the license field back to GPL-3+ in this PR and we can just close the PRs I opened (linked above)

emkll:
@redshiftzero I've updated the license fields to GPLv3, no license changes required after all. We should note that securedrop-client is licensed AGPL, while all other repos are licensed GPL3:
- securedrop-log: GPL3: https://github.com/freedomofpress/securedrop-log/blob/master/LICENSE
- securedrop-client: AGPL 3.0: https://github.com/freedomofpress/securedrop-client/blob/master/LICENSE
- securedrop-proxy: GPL 3.0: https://github.com/freedomofpress/securedrop-proxy/blob/master/LICENSE 
- securedrop-export: GPL 3.0: https://github.com/freedomofpress/securedrop-export/blob/master/LICENSE

Also opened a follow up to track packaged dependencies in https://github.com/freedomofpress/securedrop-debian-packaging/issues/125

PR #123, opened 2020-01-10
[wip] build RPM for dom0 in nightly job

redshiftzero:
Closes https://github.com/freedomofpress/securedrop-workstation/issues/357 (so that we can iterate on the RPM logic quickly)

items of note:
* right now this is running per-PR with the actual committing to the lfs repo commented out (along with all the other build jobs for speed), will fix that and take this out of draft state once the dependent PR is merged - you can see a successful build [here](https://circleci.com/gh/redshiftzero/securedrop-debian-packaging/72)
* I had to deviate from our versioning strategy used for the debian package nightlies as `-` is an invalid character, so I went with `LATEST_TAG.dev.YYMMDD.HHMMSS` which is as close as possible for consistency
* I don't plan to add a per-PR rpm build job here (only a nightly job will be added), as we have one over in securedrop-workstation, which is where the RPM build logic resides, so I see no value in duplicating that here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,13 @@ common-steps:
         git commit -m "Automated SecureDrop workstation build"
         git push origin master
 
+  - &sign_rpm
+    run: 
+      name: Sign rpms with test GPG key
+      command: |
+        echo "$SD_TEST_GPG_PRIVKEY" | base64 -d | gpg --import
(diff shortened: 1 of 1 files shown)


redshiftzero:
(closing and will reopen from a branch on this remote in a bit, since it's gotta be a branch on the remote to get access to the secrets I need)

PR #127, opened 2020-01-14
Explicitly requires hashes for pip builds

### LINKED ISSUE #79: Explicitly require hashes
The dh_virtualenv pip overrides do not specify the --require-hashes option.
https://github.com/freedomofpress/securedrop-debian-packaging/blob/948b984e9a6dce5ccbaf75ec2a8b25681441d78e/securedrop-client/debian/rules#L7

https://github.com/freedomofpress/securedrop-debian-packaging/blob/948b984e9a6dce5ccbaf75ec2a8b25681441d78e/securedrop-proxy/debian/rules#L4

Thankfully, our requirements.txt does contain hashes, and pip will default to hash checking mode if a single requirement has a hash.

We should nonetheless ensure hash checking mode is enabled to fail closed, ensuring the build logic throws an error if no hashes are specified (by setting --extra-pip-arg "--require-hashes")

conorsch:
Ensures that hashes are pinned in the requirements files used to pull in
pip dependencies for packages. Pip will already use the hashes if found,
but providing the pip option ensures that a build will fail if a reqs
file is provided that doesn't have pinned hashes.

Closes #79.

### DIFF
--- a/securedrop-client/debian/rules
+++ b/securedrop-client/debian/rules
@@ -12,6 +12,7 @@ override_dh_virtualenv:
 		--extra-pip-arg "--ignore-installed" \
 		--extra-pip-arg "--no-deps" \
 		--extra-pip-arg "--no-cache-dir" \
+		--extra-pip-arg "--require-hashes" \
 		--requirements build-requirements.txt
 
 override_dh_strip_nondeterminism:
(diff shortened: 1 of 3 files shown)


conorsch:
Closing pending more research, see https://github.com/freedomofpress/securedrop-debian-packaging/issues/79#issuecomment-573957523 for details.

PR #130, opened 2020-01-20
Rename `securedrop-workstation-svs-disp` package per VM renames

eloquence:
Ready for initial comments. Leaving in draft for now, let me know if you'd like me to try for a more history-preserving rename.

PR #131, opened 2020-01-23
Builds nightly debs for sd-svs-disp package

conorsch:
We were already building the package on PR runs, but not as nightlies.
Adds a nightly job to changes propagate to the dev lfs repo.
Identified in https://github.com/freedomofpress/securedrop-workstation/issues/282#issuecomment-577461715

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ common-steps:
 
   - &getnightlyversion
     run:
-      name: Create nightly version for debian packages
+      name: Create nightly version for python packages
       command: |
         cd ~/packaging/securedrop-*
         # Nightly versioning format is: LATEST_TAG-dev-YYMMDD-HHMMSS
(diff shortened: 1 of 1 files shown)


conorsch:
N.B. conflicts with #130

.circleci/config.yml
```diff
@@ -276,6 +276,8 @@ jobs:
       - *setsvsdispname
       - *setmetapackageversion
       - *builddebianpackage
+      - *addsshkeys
+      - *commitworkstationdebs
```
redshiftzero: heads up this job runs per-PR

per package generally we have two CI jobs, one running per-PR that builds the latest released version (see: `build-buster-securedrop-export`), and one running nightly (see: `build-nightly-buster-securedrop-export`) that bumps the version and after build commits the package

for this change we should add a new nightly job and leave this `build-buster-securedrop-workstation-svs-disp` as is

conorsch:
New build is already live via https://github.com/freedomofpress/securedrop-dev-packages-lfs/commit/43dd60f3bf6e79f57ec5c8609cbb86e0a3e3dc31, due to a mistake in an earlier version of this PR (see comment by @redshiftzero). Ready for review.

.circleci/config.yml
```diff
@@ -277,6 +277,18 @@ jobs:
...
       - *builddebianpackage
 
+  build-nightly-buster-securedrop-workstation-svs-disp:
+    docker:
+      - image: circleci/python:3.7-buster
+    steps:
+      - checkout
+      - *installdeps
+      - *setsvsdispname
+      - *setmetapackageversion
```
redshiftzero: does this bump the version? i think with this it will just keep committing versions of the same package

.circleci/config.yml
```diff
@@ -320,3 +332,4 @@ workflows:
       - build-nightly-buster-securedrop-log:
           requires:
             - build-nightly-buster-securedrop-export
+      - build-nightly-buster-securedrop-workstation-svs-disp
```
redshiftzero: let's have this `require` `build-nightly-buster-securedrop-log` (this is so the jobs don't race each other, see comment above the `nightly` key)

this will conflict with #129 but we can rebase whichever gets merged last

.circleci/config.yml
conorsch: addressed, good call

.circleci/config.yml
conorsch: Agreed. At a glance, we don't have sufficient logic to increment here, we'll need to write something new. Neither `&getlatestreleasedversion` nor `&getnightlyversion` are what we want, because both assume the package has a tag that matches its version, which isn't true since this package is managed inside this repo.

emkll:
Appended d6837be to this branch to bump metapackage version and update changelog for sd-svs-disp nightly builds, which should address the outstanding reviewer comment in https://github.com/freedomofpress/securedrop-debian-packaging/pull/131#discussion_r369903144

emkll:
Rebased on latest master (due to the merge of #129) see https://circleci.com/gh/freedomofpress/securedrop-debian-packaging/2449 for test results

PR #140, opened 2020-02-10
securedrop-workstation-config package Makefile and CI

emkll:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/424

Ensures securedrop-workstation-config can build in CI, and provide Makefile target

### Test plan
- [x] CI is passing (https://circleci.com/gh/freedomofpress/securedrop-debian-packaging/2775)
- [x] PKG_VERSION=9001 make securedrop-workstation-config correctly builds the securedrop-workstation-config metapackage
- [x] changlog updates make sense
- [x] We should still use this package
- [x] We _don't_ need nightlies for this package


securedrop-workstation-config/debian/changelog-trusty
```diff
@@ -0,0 +1,5 @@
+securedrop-workstation-config (0.1.0) unstable; urgency=medium
+
+  * Initial release.
+
+ -- SecureDrop Team <securedrop@freedom.press>  Thu, 4 Oct 2018 15:33:10 -0400
```
conorsch: Remove `changelog-trusty` altogether. We could rename it to `changelog-stretch` as a historical reference, but since we're no longer building Stretch packages, it's just cruft in the repo at this point.

securedrop-workstation-config/debian/changelog-buster
```diff
@@ -1,3 +1,9 @@
+securedrop-workstation-config (0.1.1+buster) unstable; urgency=medium
```
conorsch: Versions don't match: comment says `0.1.2`, but the string in the changelog still says `0.1.1+buster`. When I run `PKG_VERSION=0.1.2 scripts/update-changelog securedrop-workstation-config` locally, I see the changelog version properly incremented. 

Not appending a commit to clean up, since other change log items (i.e. `-trusty`) should be removed before merge.

emkll:
thanks @conorsch. Comments have been addressed, rebased on latest `master`

PR #152, opened 2020-03-11
Add securedrop-log 0.1.1 tarballs

emkll:
Build log: https://github.com/freedomofpress/build-logs/commit/e08b02a400340d0d31240962be47ef4f3ea643d4

### Suggested test plan
- [ ] Tag and signature are good, based on build logs
- [ ] Hash of tarball corresponds to build logs
- [ ] Tarball is correctly signed with release key


redshiftzero:
not strictly required anyway, but heads up we've been `git rm`ing the previous release tarball and signature just leaving the tarballs of the latest release of each project (since the previous tarballs are still in the history thanks to git-lfs)

emkll:
Woops, good catch @redshiftzero, removed securedrop-log 0.1.0 tarballs in 7a21961

PR #171, opened 2020-05-28
Add securedrop-keyring package

emkll:
Provides a securedrop-keyring package for use in all securedrop-workstation TemplateVMs/AppVMs:

1. Create a keyring package containing the Release key for production apt servers
2. Adds as dependencies to `securedrop-config` to automatically add at build-time for qubes template VMs. `whonix-gw-15` will not have the `securedrop-config` package installed.
3. Bumps securedrop--config to 0.1.3
4. The keyring package will remove the key from the default salt-provisioned location (`/etc/apt/trusted.gpg`)

see associated core PR [1]
see securedrop-workstation PR [2]

#### Test Plan

- [ ] Package builds successfully
- [ ] `make test` and test plan in [2] is completed successfully. 

[1] https://github.com/freedomofpress/securedrop/pull/5277
[2] https://github.com/freedomofpress/securedrop-workstation/pull/563/files


conorsch:
The approach here is sound, looks ready for final review to my eye, @emkll. 

> Should we provide nightlies for the securedrop-keyring package, given its infrequent update

I'd suggest we _not_ build nightlies for the keyring backup, but let's call out in standup and see if anyone can disagrees.

PR #179, opened 2020-07-16
Tweak paxctld logic in kernel metapackage

conorsch:
Prior to building the kernel modules via dkms, let's make sure that paxctld is up and running, and all flags have been applied. Otherwise, a lack of exempting flag could cause the module build to fail.

Related to https://github.com/freedomofpress/securedrop-workstation/issues/590


conorsch:
Regarding packaging procedures, I'm going to:

1. Build a new package from this PR and submit to the dev LFS repo, for inclusion on apt-test
2. Perform manual QA from apt-test repo, to determine whether the update resolves the issue
3. Once functionality is confirmed, we'll tag the "securedrop-debian-packaging" repository, rebuild, submit build logs, and open a PR to prod LFS with the new artifact.

PR #183, opened 2020-07-22
support building projects with a requirments dir

creviera:
# Description

Support building packages for projects that use a requirements directory

# Test Plan

1. Checkout `move-req-files` client branch (see https://github.com/freedomofpress/securedrop-client/pull/1128)
- [ ] Verify that you can build a client debian package when the client uses a requirements directory


kushaldas:
> Before we can merge this, all other debian package sources should also move to `requirements/` directory structure. @creviera please let me know if I misunderstood this.

Found the answer while chatting with @creviera. I will do a final round of review and hopefully merge on Monday.

creviera:
I forgot to mention that I tested both scenarios: with a req's dir and without. But forgot to add the "without" part to the test plan. That'll be a good check to add to your final review to gain more confidence in this change. Thanks for the review so far!

Issue #182, opened 2020-07-22
Make build more reproducible


creviera:
# Description 

Right now, to build a workstation package, we run

```
PKG_PATH=path/to/tarball PKG_VERSION=x.y.z make some-package
```

If we make the tarballs reproducible, then we could just run:

```
PKG_VERSION=x.y.z make some-package
```

Which would also allow us to skip manual tarball checksum verification and signing. We still have the python wheels left over as the remaining piece that is not reproducible but as @emkll put it earlier today, we're working our way down the chain of reproducibility. Curious what others know and think about this.

eloquence:
@conorsch has expressed interest in doing a first investigation of tarball reproducibility during the 7/23-8/5 sprint.

FWIW, some ideas for reproducible/idempotent tarballs here: https://stackoverflow.com/a/54908072 (@rmol also mentioned using `--mtime`, which per that comment may not be fully sufficient).

conorsch:
> Which would also allow us to skip manual tarball checksum verification and signing. 

Good point. Zooming out a bit, we've got two fundamental goals with the build logic:

1. Be absolutely certain that the code we're shipping is the code we reviewed and tagged with the prod key.
2. Enable anyone, SD maintainer or not, to rebuild the _exact same_ `.deb` files to confirm that the binary debs we ship match what's in our open source repos.

As @creviera mentions, the tarballs (created via `python setup.py sdist` in the various project repos) are not yet fully reproducible, mostly due to a small bit of variation in the metadata. Fortunately, support for the `SOURCE_DATE_EPOCH` env var are coming to a setup.py near you: https://github.com/python/cpython/pull/20331 — but that doesn't help us _today_. Because the tarballs aren't reproducible, we're signing them and committing them to this repository, then we're also setting SOURCE_DATE_EPOCH to a static value for the debian packages:  https://github.com/freedomofpress/securedrop-debian-packaging/blob/f6879bfe320158ef8e0e83323f12733dffce3462/scripts/build-debianpackage#L88-L89

Since we already have a method for ensuring reproducible builds of the binary debian packages, we technically don't need fully reproducible tarballs in order to satisfy 2 above. So, we can remove some of the manual steps that @creviera rightly points out aren't adding much, running simply:

```
# Current method
PKG_PATH=path/to/tarball PKG_VERSION=x.y.z make some-package

# Potential new method
make some-package
```

[...]

PR #185, opened 2020-07-28
Builds source tarballs dynamically

conorsch:
Ready for review. For reference, see discussion in https://github.com/freedomofpress/securedrop-debian-packaging/issues/182#issuecomment-664709727

## Changes
Full backwards compatibility is retained, so the original invocation will work just fine:

```
PKG_VERSION=0.2.1 PKG_PATH=tarballs/securedrop-client-0.2.1.tar.gz make securedrop-client
```

There's now an optional, terser step to build a package, where the version defaults to the latest tag, and the tarball is generated dynamically from that tag:

```
make securedrop-client
```

Regardless of which method is used, the `.deb` file generated from it is byte-for-byte identical over multiple runs. In both cases, the timestamp information is pulled in from the package changelog. Ideally we'd be using SOURCE_DATE_EPOCH in all places to manage the timestamp info, but not all tools support it yet (particularly not 'python setup.py sdist'). Hopefully we'll see movement upstream soon, and then we can standardize on using SOURCE_DATE_EPOCH everywhere to control timestamps. 

It's possible we'll want to break out the logic that chains python-sdist -> git -> tar -> gz so we can reuse it elsewhere, but not doing that now. 

### debs are reproducible
Regardless of how the tarball was generated, or even which was used, the resulting `.deb` file for a specific package version should be identical. Confirm that by running the following:

```
#!/bin/bash
set -e

results_dir=/tmp/test_results
rm -rf "$results_dir"
mkdir -p "$results_dir"

echo "Building from new dynamic logic"
make securedrop-client
cp /home/user/debbuild/packaging/securedrop-client_0.2.1+buster_all.deb "${results_dir}/dynamic.deb"

echo "Building from historical static tarball"
PKG_VERSION=0.2.1 PKG_PATH=tarballs/securedrop-client-0.2.1.tar.gz make securedrop-client
cp /home/user/debbuild/packaging/securedrop-client_0.2.1+buster_all.deb "${results_dir}/static.deb"

echo "Comparing results"
sha256sum ${results_dir}/*
diffoscope ${results_dir}/* && echo "SUCCESS: fully reproducible"
```

You should see output of `4b65f8a91f8555a2e0303ce2cdb97a58addddbc7a2e6d3567805b8aa0a0f4ed3`, with the "SUCCESS" message printed at the bottom.

### tarballs are reproducible

The tarballs are an _intermediate_ artifact in our build process. Let's confirm that if we build them dynamically from source, we're getting identical artifacts—which is a new feature introduced by this PR.

```
# Prepare a results dir
rm -rf /tmp/results
mkdir /tmp/results

# Use the new dynamic build logic
make securedrop-client
cp /tmp/securedrop-client/dist/*.tar.gz /tmp/results/1.tar.gz
rm -rf /tmp/securedrop-client

# Run it again
make securedrop-client
cp /tmp/securedrop-client/dist/*.tar.gz /tmp/results/2.tar.gz
rm -rf /tmp/securedrop-client

# Compare the results!
sha256sum /tmp/results/*.tar.gz
diffoscope /tmp/results/*.tar.gz && echo "SUCCESS: fully reproducible
```

### DIFF
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -27,6 +27,8 @@ TOP_BUILDDIR="$HOME/debbuild/packaging"
 mkdir -p "$TOP_BUILDDIR"
 rm -rf "${TOP_BUILDDIR:?}/${PKG_NAME}"
 mkdir -p "${TOP_BUILDDIR}/${PKG_NAME}"
+# Move changelog into place (we have separate changelogs for each platform)
+PLATFORM="$(lsb_release -sc)"
 
 # Validate required args.
 if [[ -z "${PKG_NAME:-}" ]]; then
@@ -34,9 +36,23 @@ if [[ -z "${PKG_NAME:-}" ]]; then
     exit 1
 fi
 
+
+function find_latest_version() {
+    repo_url="https://github.com/freedomofpress/${PKG_NAME}/releases"
+    curl -s "$repo_url" \
+        | perl -nE '$_ =~ m#/releases/tag/(v?[\d\.]+)\"# and say $1' \
+        | head -n 1
+}
+
 if [[ -z "${PKG_VERSION:-}" ]]; then
-    echo "Set PKG_VERSION of the build";
-    exit 1
+    echo "PKG_VERSION not set, inferring from recent releases..."
+    PKG_VERSION="$(find_latest_version)"
+    if [[ -z "$PKG_VERSION" ]]; then
+        echo "Failed to infer version"
+        exit 1
+    else
+        echo "Using PKG_VERSION: $PKG_VERSION"
+    fi
 fi
 
 # Copy over the debian directory (including new changelog) from repo
@@ -80,8 +96,6 @@ fi
 
 printf "Building package '%s' from version '%s'...\\n" "$PKG_NAME" "$PKG_VERSION"
 
-# Move changelog into place (we have separate changelogs for each platform)
-PLATFORM="$(lsb_release -sc)"
 echo "$TOP_BUILDDIR/$PKG_NAME/"
 mv "$TOP_BUILDDIR/$PKG_NAME/debian/changelog-$PLATFORM" "$TOP_BUILDDIR/$PKG_NAME/debian/changelog"
 
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -58,14 +58,26 @@ fi
 # Copy over the debian directory (including new changelog) from repo
 cp -r "$CUR_DIR/$PKG_NAME/" "$TOP_BUILDDIR/"
 
+function build_source_tarball() {
+    repo_url="https://github.com/freedomofpress/${PKG_NAME}"
+    build_dir="/tmp/${PKG_NAME}"
+    rm -rf "$build_dir"
+    git clone "$repo_url" "$build_dir"
+    git -C "$build_dir" tag --verify "$PKG_VERSION" 1>&2
+    git -C "$build_dir" checkout "$PKG_VERSION" 1>&2
+    (cd "$build_dir" && python setup.py sdist 1>&2)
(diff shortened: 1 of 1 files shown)


scripts/build-debianpackage
```diff
@@ -27,29 +27,57 @@ TOP_BUILDDIR="$HOME/debbuild/packaging"
...
+PLATFORM="$(lsb_release -sc)"
 
 # Validate required args.
 if [[ -z "${PKG_NAME:-}" ]]; then
     echo "Set PKG_NAME of the build";
     exit 1
 fi
 
+
+function find_latest_version() {
```
creviera, line 40: this is a good idea. will your change still support supplied version numbers in case we're building a dev release?

conorsch: Yes, fully backwards compatible in that PKG_VERSION & PKG_PATH can still be provided as overrides, they're just not _required_. The verification logic needs some work before it's ready for full review—at least:

1. use a dedicated keyring so that _only_ the prod signing key is used to validate tags
2. detect if an "rc" tag is being used, and if so, don't expect a prod sig

Then it should be ready to bang around on.

scripts/build-debianpackage
```diff
@@ -27,29 +27,90 @@ TOP_BUILDDIR="$HOME/debbuild/packaging"
...
 
+# Ensures that a given git tag is signed with the prod release key
+# If "rc" is in the tag name, this will fail.
+function verify_git_tag() {
+    local d
+    local t
+    d="$1"
+    t="$2"
+    prod_fingerprint="22245C81E3BAEB4138B36061310F561200F4AD77"
+    git -C "$build_dir" tag --verify "$PKG_VERSION" 2>&1 \
```
kushaldas, line 70: Can we please add some error messages here somehow?  At least for missing key or expired key :)
I was wondering what is wrong for some time with the following output. 

```
✦ ❯ make securedrop-client
		PKG_NAME="securedrop-client" ./scripts/build-debianpackage
	gpg: key 310F561200F4AD77: public key "SecureDrop Release Signing Key" imported
	gpg: Total number processed: 1
	gpg:               imported: 1
	gpg: Signature made Tuesday 28 April 2020 07:57:53 PM IST
	gpg:                using RSA key 22245C81E3BAEB4138B36061310F561200F4AD77
	gpg: Good signature from "SecureDrop Release Signing Key" [unknown]
	gpg:                 aka "SecureDrop Release Signing Key <securedrop-release-key@freedom.press>" [unknown]
	gpg: WARNING: This key is not certified with a trusted signature!
	gpg:          There is no indication that the signature belongs to the owner.
	Primary key fingerprint: 2224 5C81 E3BA EB41 38B3  6061 310F 5612 00F4 AD77
	Checking that SHA256SUMs from mirror match signed file...  OK
	PKG_VERSION not set, inferring from recent releases...
	Using PKG_VERSION: 0.2.1
	securedrop-client is a Python package
	PKG_PATH not set, building from source (version 0.2.1)...
	Cloning into '/tmp/securedrop-client'...
	remote: Enumerating objects: 64, done.
	remote: Counting objects: 100% (64/64), done.
	remote: Compressing objects: 100% (48/48), done.
	remote: Total 9276 (delta 32), reused 26 (delta 16), pack-reused 9212
	Receiving objects: 100% (9276/9276), 6.95 MiB | 2.51 MiB/s, done.
	Resolving deltas: 100% (6863/6863), done.
	make: *** [Makefile:9: securedrop-client] Error 1
```

conorsch: Great point, @kushaldas. Added. Try again and confirm that an error message displays for you.

kushaldas (approved):
- [x] Deb packages are reproducible
- [x] tarballs are reproducible

This is amazing, approved. I will still open a separate issue for a better error messages related to GPG.

PR #190, opened 2020-08-28
Provide mime and paxctld configuration for template consolidation

### LINKED ISSUE #188: Consolidate MIME type configuration in `securedrop-workstation-config` package
Currently, MIME types are provisioned via Salt and via packages (i.e. `securedrop-client`, `securedrop-export` and `securedrop-workstation-svs-disp`). As part of the template consolidation epic, we want to provide a set of MIME type configurations in the `securedrop-workstation-config` package, which would be installed in the TemplateVMs.

The package should contain:
- in `/opt` : all possible mimetype definitions for all AppVMs
- in `/usr/share/applications`: a “defensive” mimetype definition for to be installed in a template (default action for all mimetypes is to open in disposable VM)

This can be done initially in an additive way, without removing the existing MIME type handling logic.

emkll:
Closes #188 , test plan can be found in TBD

* Provides a standard paxctld configuration for all Workstation VMs
* Provides (but does not enforce) mime type associations for sd-app, sd-devices, and sd-viewer

Due to order-of-operations, we will need to add the defensive mimetype configuration _after_ the clean up and consolidation, as `securedrop-client`, `securedrop-export` and `securedrop-workstation-svs-disp` are competing with `/usr/share/applications/mimeapps.list`, and we do not want to introduce breakage of the functionality provided by securedrop-export` (I propose we open a follow-up issue)


creviera (approved):
LGTM

PR #191, opened 2020-08-31
Adds script to generate tag message

conorsch:
When tagging the "securedrop-debian-packaging" repository, we include
a list of all package versions, to make reasoning about releases easier
down the road. Collecting this information manually is error-prone, so
let's provide a routine method of generating that info so it can easily
be pasted into a prompt for an annotated git tag, consistent with docs.

### DIFF
--- a/scripts/generate-tag-message
+++ b/scripts/generate-tag-message
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Utility script to generate a signed tag for the packaging repo:
+#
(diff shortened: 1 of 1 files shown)


eloquence:
Discussed in sprint planning today, can wait until the next package release.

conorsch:
Rebased, mostly to include #198, but also because we're moving into QA, and this script will likely save maintainers some time.

emkll:
rebased on latest `main`

emkll (approved):

PR #194, opened 2020-09-08
Update README intro per current state

eloquence:
In particular, `apt-test-qubes.freedom.press` is no longer used.

Also clarifies which repo the release tag is pushed to.


emkll (approved):
:+1:

PR #198, opened 2020-10-05
Reorganizes mimetype declarations for template consolidation

conorsch:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/471

Closes #197 

Includes updates to packages for the Workstation, in order to achieve template consolidation. These packages are currently being served via a "template-consolidation" channel on apt-test: https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/65 Specifically, the changes presented here are:

* New package `securedrop-workstation-viewer`, superseding `securedrop-workstation-svs-disp`, thereby closing #197 
* Updated `securedrop-proxy` package, adjusting the config filepath, built from https://github.com/freedomofpress/securedrop-proxy/pull/79
* Updated `securedrop-export` & `securedrop-client` to remove mimetype associations (building on work from #190)
* Updates `securedrop-workstation-config` to provide default mimetype assocations for `open-in-dvm`

## Testing

Check out https://github.com/freedomofpress/securedrop-workstation/pull/619, which will configure both of these new packages via a custom apt-test channel. There's also a detailed test plan at https://github.com/freedomofpress/securedrop-workstation/wiki/0.5.0-Test-Plan

PR #206, opened 2020-10-27
Removes mimetypes from securedrop-client package

conorsch:
Follow up to #198. Omitted in the final version for review.


emkll (approved):
LGTM, looks like it will make CI green for https://github.com/freedomofpress/securedrop-export/pull/65/files and in line with our template consolidation changes.

PR #211, opened 2020-12-18
Makes wheel builds reproducible, with tests

conorsch:
### Overview
Refs #196

Makes two changes to our existing wheel build process:

  1. Setting SOURCE_DATE_EPOCH env var
  2. Setting a hardcoded dir path for all wheels builds

With those changes, the `./scripts/build-sync-wheels` script builds wheels that are byte-for-byte reproducible. I also included a test-only Makefile target that was handy during development. It's not yet ready for CI, because I'm seeing the platform vary in the reprotest build environment. More input appreciated if anyone can resolve that problem, see initial report in https://github.com/freedomofpress/securedrop-debian-packaging/issues/196#issuecomment-733417904

### Testing 
First, it's helpful to observe the wheel checksums changing over multiple builds, to understand the problem we're trying to solve.

1. Check out a new branch, _based on main_, rather than this feature branch.
2. Run `git rm localwheels/*.whl` and commit.
3. Run `./build-sync-wheels -p ../securedrop-client`, updating the dirpath as necessary. Commit the results.
4. Run `git rm localwheels/*.whl`, but don't commit the results.
5.  Run `./build-sync-wheels -p ../securedrop-client` again, and commit the results.
6. Run `git show --name-status` and confirm that many of the `localwheels/*.whl` files changed. 

That's a convoluted workflow because the build-sync-wheels script will politely avoid overwriting wheel files that already exist. In this PR, a `--clobber` flag is added to the script to force overwrites, making comparison easier.

OK, now let's verify the new behavior! 

1. Check out a new branch, _based on this branch_. Note that the `localwheels/*.whl` files are identical to those on the main branch.
2. Run `./build-sync-wheels --clobber -p ../securedrop-client`, updating the dirpath as necessary. Commit the results.
3. Run `./build-sync-wheels --clobber -p ../securedrop-client` again 
4. Run `git status`, confirm that there are no local modifications. The wheels were fetched, but identical. 

That's about it. Eventually, I'd love to drop the requirement that we store the wheels in version control, separately from the deb packages that contain them. But the work here is a step in that direction: once we have reproducible wheels, we can work on skipping artifact storage as long as the final output is deterministic.

PR #213, opened 2021-01-05
Rebuild all wheels reproducibly and use wheels from local filesystem

### LINKED ISSUE #212: Rebuild all the current wheels in a reproducible way & use them for package building
As we have https://github.com/freedomofpress/securedrop-debian-packaging/pull/211 merged now, we should rebuild all the wheels for the following packages. 

- `securedrop-client`
- `securedrop-log`
- `securdrop-proxy`

The same PR should also change the package building rules to start using the localwheels directory.

kushaldas:
This patch stops accessing the webbased index for the wheels,
instead uses the local filesystem path.

Fixes #212 (work remaining) 

## Next step in this PR.

- [x] Someone with signing capability should regenerate all the wheels for `securedrop-client`, `securedrop-proxy` and `securedrop-log` projects.  `./scripts/build-sync-wheels --clobber -p ../securedrop-client` is one example.
- [ ] Then we should regenerate the `build-requirements.txt` files for all the 3 Debian packages
- [ ] CI should be green
- [ ] Remove our python package index.(temporarily via infra work) 
- [ ] CI still should be green.

## How to test?

- [ ] `make securedrop-client`
- [ ] `make securedrop-proxy`
- [ ] `make securedrop-log`
- [ ] CI should be green.

All of the above commands should pass.


conorsch:
> regenerate all the wheels

@kushaldas I've appended new wheel files and checksums here, and marked the task as complete in your list. Back over to you!

eloquence:
(Retitled for clarity, please don't hesitate to edit if I misunderstood but I wanted to make it clearer that this PR also updates all the existing wheel files.)

kushaldas:
Next, we will have to merge these 3 PRs:

- [ ] https://github.com/freedomofpress/securedrop-proxy/pull/81
- [ ] https://github.com/freedomofpress/securedrop-client/pull/1203
- [ ] https://github.com/freedomofpress/securedrop-log/pull/21

Only after those are merged, we will be able to see CI progressing for this PR.

conorsch:
@kushaldas I haven't provided in-depth review of all the checksums in the component PRs you link to, but plan to do so tomorrow. While testing that, I might append yet another test-only commit onto this PR, tracking the feature branches in those repos, to confirm everything is green prior to approving.

Thanks for preparing! Looking forward to landing these changes.

kushaldas:
So the package build tests were failing as they were building the last release. And the repository now has different wheels from the last release. In my temporary commit, I am doing night build instead of rebuilding the last release. 

I am still not sure how to fix the `reproducibility-checks` test for this purpose. @conorsch please let me know what do you think.

conorsch:
>  And the repository now has different wheels from the last release. 

From the last release of which package? Are you recommending that the wheels in this PR be regenerated to resolve?

kushaldas:
So the package build tests were failing as they were building the last release. And the repository now has different wheels from the last release. In my temporary commit, I am doing night build instead of rebuilding the last release. 

I am still not sure how to fix the `reproducibility-checks` test for this purpose. @conorsch please let me know what do you think.

> > And the repository now has different wheels from the last release.
> 
> From the last release of which package? Are you recommending that the wheels in this PR be regenerated to resolve?

- securedrop-client
- securedrop-proxy
- securedrop-log

Generally the CI is trying to build the last release of these packages, which is failing because now we have new wheels. My temporary commit tries to build the nightly, that is why is passing for those builds. But, the `reproducibility-checks` test tries to build from the old tarballs, which contains the old wheel checkcums. This is why <https://circleci.com/gh/freedomofpress/securedrop-debian-packaging/6586?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link> is still failing.

emkll:
> I am still not sure how to fix the reproducibility-checks test for this purpose.

@kushaldas you correctly identify that the reproducibility check rely on the archived tarballs[1], which were have considered deprecating. It sounds like we could replace this step by manually cloning and building the tarball locally on the latest release tag on `main` branch of the target repos using some of the logic introduced in [2]. You can then, for the purposes of this PR, add a temporary commit to check out the feature branch with the updated hashes.
[1] https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/.circleci/config.yml#L484
[2] https://github.com/freedomofpress/securedrop-debian-packaging/pull/185

sha256sums.txt
```diff
@@ -1,38 +1,39 @@
...
-1855003e64dcb934556ad79994ba1a3a852da337e353d84d3b4ef75031913451  furl-2.0.0-py2.py3-none-any.whl
+9f50360f6e4a0f1d0a35fb4997878e7186a73331f0fde5f6fc9b1bb9f006e6cc  furl-2.0.0-py2.py3-none-any.whl
 fdcaedc1fb19a63d7d875b0105b0a5b496dd0989330d454a42bcb401fa5454ec  furl-2.0.0.tar.gz
-491f674364ba3232ed1eb4c1eb7407887f62cef6c300aad7df6e01acd88ffb25  idna-2.7-py2.py3-none-any.whl
+a967a752dba5492be34f30f8633cee512d2ec923fe3a2e201d87353b540d2955  idna-2.7-py2.py3-none-any.whl
 684a38a6f903c1d71d6d5fac066b58d7768af4de2b832e426ec79c30daa94a16  idna-2.7.tar.gz
-614c22fe1a5b0a3f46f6c5c43ff2e6795e4e784328d559ec9dc49db0f06b3a75  Mako-1.0.7-py3-none-any.whl
+99d041a616a3655725dffe459916627b75640a7d045ed33f51ce158a168ca3ef  Mako-1.0.7-py3-none-any.whl
 4e02fde57bd4abb5ec400181e4c314f56ac3e49ba4fb8b0d50bba18cb27d25ae  Mako-1.0.7.tar.gz
 c6b726d2e9d6300a044cf6a37627f10994268d6ac39464bc0d725126609311a5  MarkupSafe-1.0-py3-none-any.whl
```
creviera, line 15: can this be removed?

kushaldas: > can 1.0 entries be removed?

I think we should do it later with a new PR as we mess up, it will difficult to figure out where it broke (if it is the new work or removing old things).

kushaldas: Oh, I guess you are talking about removing the old MarkupSafe wheel removal, that can be done.

kushaldas:
I could not do it via packages/tarballs from current `main` of those packages, as I will have to remove the signing check and all related checks etc. Can anyone please try to write it in a better way?

conorsch:
I haven't taken a look at this yet, @kushaldas, but will do so later in the week.

kushaldas:
@conorsch I tried it in https://github.com/freedomofpress/securedrop-debian-packaging/pull/214 and I think finally I got it properly.
There I am rebuilding the tarball for `securedrop-proxy` and `securedrop-log` and disable gpg signature. We can enable gpg signature after the next release of those packages.

conorsch:
@kushaldas Nice work in #214! I took the liberty of appending yet more partial work, specifically using `reprotest` to verify that the checksums matched on the `.deb` files, same as on the `.whl` files (#211). As part of that work, I pulled in your test-only commits into this PR, and I've also re-broken the tests. As part of final review, let's make sure to drop all the test-only commits. 

Based on the new changes, I have a couple immediate observations:

1. The `securedrop-export` package needs its own PR, since it's still building from the pypi.securedrop.org/ index on this branch
2. reprotest _requires_ that build artifacts be relative to the project directory. That's already true of the whls, but the debs we've been placing in `~/debbuild/packaging/`, and I propose we move that logic to `./build/debbuild/packaging/` in this PR. I've tried to make that migration here, but it's not passing in CI yet.
3. Assuming `make test` passes for you locally, and in CI, it looks like we can _remove_ the following CI checks:
  - build-buster-securedrop-client
  - build-buster-securedrop-proxy
  - build-buster-securedrop-log
  - build-buster-securedrop-export (after we have the checksums updated, mentioned above)

I'm suggesting removal of those tests on every PR run, because the reprotest suite effectively verifies that the build passes—and then passes again! If you agree, please take a stab at adjusting the settings. I'd also appreciate your help on getting the new reprotest suite passing in CI. You'll see that I've moved it from a machine to container executor in CI, but that itself isn't enough. See what you can figure out. We're almost there! :rocket:

conorsch:
Took another look at this today. Looks like we're running into a CI-specific problem with `setarch`, just as we saw in https://github.com/freedomofpress/securedrop-debian-packaging/pull/211/commits/e29360673066aebcaf6491e72c608111dceddfe1. It appears this problem prevents us from using `reprotest` to verify the deb package builds, because we can _either_ use Focal VM images, which cause the wheel lookups to fail, since Focal uses python3.8 rather than Buster's 3.7, or we can use a Buster container image, which will fail on setarch. I haven't found a way to tell reprotest to skip the `setarch` line, so I believe we're blocked on that for now. 

I'll work on adjusting the tests to accommodate. Overall, the logic here is very much what we want! Will poke at reprotest in CircleCI containers another day to understand the root cause better.

emkll:
@conorsch @kushaldas have you tried the "next gen" circle images? They have python buster images that can be used here:
https://circleci.com/docs/2.0/circleci-images/#python

We have recently switched to these for some core CI targets: https://github.com/freedomofpress/securedrop/pull/5720

conorsch:
No, but worth a shot! Thanks for mentioning.

conorsch:
> The securedrop-export package needs its own PR, since it's still building from the pypi.securedrop.org/ index on this branch

Fixed in this branch. A clarification, though: yes, the build logic was erroneously using `pypi.securedrop.org`, but the securedrop-export package doesn't have any external dependencies, so that wasn't breaking the build. Therefore no additional PR is required. Updated the docs to remove mention of the pypi.securedrop.org URL entirely, since we plan to remove it in the future.

> have you tried the "next gen" circle images?

I tried those, but to no avail. The problem appears to be that in CircleCI, containers are run with reduced privileges, prohibiting the use of the `setarch` call. Unfortunately reprotest doesn't allow us to disable that call via configuration, so I patched the file only in the CI env: https://github.com/freedomofpress/securedrop-debian-packaging/pull/213/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R306 Not the prettiest solution, but it lets us remove the manual checksum comparison specific to the CircleCI testing logic. The reproducibility checks can now be run locally, in addition to CI, which should help greatly in future debugging sessions.

CI is now passing across the board, so I'm marking ready and proceeding with final review.

conorsch:
Merged the associated component PRs, then reverted the temporary commits on this branch. See fully passing build here, with temporary commits reverted: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-debian-packaging/885/workflows/3fb0f623-77a7-4b16-9c9b-2497e8866f5c I'll rebase again to drop those temporary commits entirely, confirm passing, then merge.

conorsch (approved):

conorsch:
After merging all associated PRs, including this one, into the main branches for all repos, I re-ran the tests in CircleCI, and they're showing green across the board:

* `securedrop-debian-packaging` - https://app.circleci.com/pipelines/github/freedomofpress/securedrop-debian-packaging/887/workflows/bffffdb9-7562-41e5-bd11-029804c1249f
* `securedrop-client` - https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/548/workflows/67d81d1a-9b59-41e0-b9bf-6fbd1097ab83
* `securedrop-log` - https://app.circleci.com/pipelines/github/freedomofpress/securedrop-log/62/workflows/50bf7ed8-9039-41a8-84fe-447174c4d87c
* `securedrop-proxy` - https://app.circleci.com/pipelines/github/freedomofpress/securedrop-proxy/26/workflows/89e74617-2199-4aae-baee-ac354f7f90e6

There may be a few PRs, especially in the securedrop-client repo, that will need to be rebased to keep tests passing. Other than that, no further action required.

PR #221, opened 2021-03-10
 Improvements to xdg-open configuration

eloquence:
## Description

- Fixes a syntax error in `.desktop` file that could cause `xdg-open` to Fail Open in limited circumstances
- Hardens mailcap rules to enforce Fail Closed behavior for `xdg-open` invocations
- Adds .desktop file linter to package dependencies and a Makefile target to run it (will add to CI in a follow-up PR)

## Status

Ready for review

## Test plan

1. Run `apt-cache policy desktop-file-utils` in your dev VM.
2. If the package is installed, remove it with `sudo apt-get remove desktop-file-utils`.
3. Check out the code in this branch
4. Run `make install-deps` in the `securedrop-debian-packaging` directory.
5. Repeat step 1.
6. - [ ] Observe that `desktop-file-utils` is now installed
7. Run `make lint-desktop-files`.
8. - [ ] Observe that it exits with exit code 0 (`echo $?`).
9. Build a new `securedrop-workstation-config` package with `PKG_VERSION=0.1.6 make securedrop-workstation-config`. Locate the built package (on my system: `/home/user/securedrop-debian-packaging/build/debbuild/packaging/securedrop-workstation-config_0.1.6+buster_all.deb`) and copy it to your `sd-large-buster-template` and `sd-small-buster-template`. Due to the RPC policy, I find it easiest to pull it into `dom0` first, and then use `qvm-copy-to-vm`.
10. From the relevant `~/QubesIncoming` directory in `sd-small-buster-template` and `sd-large-buster-template`, install the new package via `dpkg --install securedrop-workstation-config_0.1.6+buster_all.deb`.
11. - [ ] Observe that the config package is being installed over the previous version.
12. Power off both template VMs (`sudo poweroff`) and restart all running SecureDrop Workstation AppVMs (to pick up the template update).
13. In your dev VM, update your `securedrop-workstation` checkout to the `config-hardening` branch, run  `make clone` in `dom0`, ensure configuration and private key are present, and update your installation with `make staging`. Make a coffee.
14. Open a terminal in `sd-app`
15. Create an ODT file: `touch a.odt`
16. Open the file via `qvm-open-in-vm sd-devices a.odt`
17. - [ ] Observe that the test file you created is opened in a disposable VM.
18. Run the test suite in `dom0` via `make test`.
19. - [ ] Observe that all tests are passing (or any failures are due to local configuration discrepancies), including the newly added `test_mailcap_hardened` tests.
[...]

PR #223, opened 2021-03-10
Add .desktop file linting to CI

eloquence:
# Description

To catch syntax errors like the one fixed in #221, this PR adds the existing `lint-deskop-files` Makefile target to CI. I've opted to combine it with the test job for now, since the `install-deps` step is pretty heavy and it's only this one linter for now; would suggest we reorganize if/when we add more linters.

# Status

Ready for review


conorsch (approved):
I like it!

PR #237, opened 2021-03-30
Fixes #232 Uses python3 setuptools for sdist

### LINKED ISSUE #232: had to install python-setuptools
## Description

Recently when building debs, I had to install `python-setuptools` in order for this command to work:
```py
PKG_VERSION=0.4.1 make securedrop-client
```

It looks like we install `python3-setuptools` via `make build-deps` but it appears the script to build debian packages is run using system python, which is why I had to install the python2 version.

kushaldas:
We are now using the Python3 setuptools package
to create the source tarballs.

Fixes #232

## How to test?

- [x] make sure that no `python-setuptools` is installed in the system
- [x] `PKG_VERSION=0.4.1 make securedrop-client` should be able build the package following the standard build procedure.


kushaldas:
I am wondering why is it trying to build the `dom0-rpm`, maybe I am too tired to see.

emkll:
CI is likely failing here de due to changes in the dom0 build logic introduced in https://github.com/freedomofpress/securedrop-workstation/pull/666

emkll (approved):
Changes look good to me, I've appended a commit to remove the dom0 build logic from this repo.
If we want to reinstate dom0 builds, we can revert 42a185fb22cbb4171e0e6695634c5f29a2cd1158 and adjust the rpm signature script to work in Debian

Issue #218, opened 2021-02-12
 pip-21.0.1 is breaking reproducible wheels


kushaldas:
The `--build` is now deprecated. Tracking it in the upstream issue https://github.com/pypa/pip/issues/9604  

Note: we should add a monthly CI job to to test reproducibility using the latest `setuptools` and `pip`.

emkll:
Thanks for opening the issue @kushaldas . Right now, the `reprotest-wheels` CI job is _not_ failing against latest main , so we wouldn't have caught it with simply adding a monthly CI job as-is: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-debian-packaging/902/workflows/09def272-fb86-4906-a63a-53a9f6de4c60/jobs/6978

kushaldas:
> Thanks for opening the issue @kushaldas . Right now, the `reprotest-wheels` CI job is _not_ failing against latest main , so we wouldn't have caught it with simply adding a monthly CI job as-is: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-debian-packaging/902/workflows/09def272-fb86-4906-a63a-53a9f6de4c60/jobs/6978

Yes, that is why I mentioned to use **latest** `setuptools` and `pip` in that monthly job.

pradyunsg:
```sh-session
$ python -m build --wheel ./simplejson-3.17.2 --no-build-isolation
...
$ shasum -a 256 ./simplejson-3.17.2/dist/simplejson-3.17.2-cp38-cp38-macosx_10_9_x86_64.whl
6bbb49faf3a94233d3e1dd5059040b939169f2485426cbbef97dfbfa0ecda1ee  ./simplejson-3.17.2/dist/simplejson-3.17.2-cp38-cp38-macosx_10_9_x86_64.whl
$ python -m build --wheel ./simplejson-3.17.2 --no-build-isolation
...
$ shasum -a 256 ./simplejson-3.17.2/dist/simplejson-3.17.2-cp38-cp38-macosx_10_9_x86_64.whl
6bbb49faf3a94233d3e1dd5059040b939169f2485426cbbef97dfbfa0ecda1ee  ./simplejson-3.17.2/dist/simplejson-3.17.2-cp38-cp38-macosx_10_9_x86_64.whl
$ python -m build --wheel ./simplejson-3.17.2
...
[...]

PR #238, opened 2021-03-31
Adds bootstrapping wheels for the `build` tool

kushaldas:
# How to test?

Remember to put your OpenPGP public key in the pubkeys/ directory as maintainer_name.pub. Right now we have the release key, and @kushaldas's key there.

In future we can create another shell script for the whole bootstrapping process using the commands below.

```
python3 -m venv .venv
source .venv/bin/activate
python3 -m pip install pip-tools
pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
python3 -m pip install -r requirements.txt
./scripts/build-sync-wheels --cache ./bootstrap -p $PWD
BOOTSTRAP=true ./scripts/sync-sha256sums
gpg --armor --output bootstrap-sha256sums.txt.asc --detach-sig  bootstrap-sha256sums.txt
BOOTSTRAP=true ./scripts/verify-sha256sum-signature
# Now update the build-requirements.txt
PKG_DIR=$PWD BOOTSTRAP=true ./scripts/update-requirements
```

- [ ] `python3 -m pytest -vv tests/test_reproducible_wheels.py`

## What is happening in the test?

In the test we are actually installing our bootstrapped wheels into the virtualenv, and then using the existing scripts to build the wheels.  You can see the same in the readme of the branch https://github.com/freedomofpress/securedrop-debian-packaging/tree/build_with_build#0-enable-the-virtualenv

## Points to remember

- `idna` project is getting extracted with different permissions than our previous wheels. Means different `sha256sum` than our existing wheel.
- All extension based wheels will have a different `sha256sums` than our existing wheels.


kushaldas:
Just now noticed that the `reprotest-wheels` CI step is running on Ubuntu, but all of our wheels are for Python3.7 on Debian Buster. 
https://github.com/freedomofpress/securedrop-debian-packaging/blob/e960cd988f4687d49580b80fb12262958330a423/.circleci/config.yml#L248

I will push another time to update this. The current test failure is happening due to mismatch of the `sha256sum` of the bootstrap wheels, which are not matching to what I pushed from my local system. I will dig more.

https://circleci.com/api/v1.1/project/github/freedomofpress/securedrop-debian-packaging/7713/output/102/0?file=true&allocation-id=60647230441cc96504afc970-0-build%2F389C71BD

zenmonkeykstop:
with `git-lfs` installed first, most dependencies match, except for Cython:

```
...
Processing ./bootstrap/zipp-3.4.1-py3-none-any.whl
Processing ./bootstrap/pip-21.0.1-py3-none-any.whl
Processing ./bootstrap/setuptools-54.0.0-py3-none-any.whl
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    cython==0.29.22 from file:///home/circleci/project/bootstrap/Cython-0.29.22.tar.gz (from -r build-requirements.txt (line 2)):
        Expected sha256 8c9eec7e9de2a30861ca347d0a149cc1482de12fc765fa06c414930e8ce20d0a
             Got        df6b83c7a6d1d967ea89a2903e4a931377634a297459652e4551734c48195406
```
But! I think this is legitimate, the expected hash is for the .whl and the returned hash is actually correct for the tarball on a local checkout.

conorsch:
Took an initial spin through. Solid work here. My questions at present are mostly around maintanance tasks.

Why not store the new wheels for `build` and its dependencies inside the `localwheels/` directory? We already have a directory for wheels we've built ourselves, but we're not using it in this PR. Conditional if/else logic has been added to most scripts, to support targeting either `bootstrap/` or `localwheels/`, but it seems they'd work just as well unchanged, if the wheels were added to `localwheels/`.
 
 What is `build-requirements.txt` used for? It was added in this PR, but I don't see it referenced in any of the test plans, or in the scripts. As of this PR, this repo now has three (3) Python requirements files. Coupled with the detached sig logic re-confirming the same SHA256 checksums as in the various component repos' build-requirements files, that's a lot of machinery to maintain. If we can remove `build-requirements.txt`, then we're back down to two (2), same as on the `main` branch.
 
 Based on prior discussion, mostly in #218, the benefits of these changes appear to be:
 
 - we can ensure that the `localwheels/*.whl` files are built reproducibly, even when pip upgrades introduce breaking changes
 - we can potentially build the packages offline, given that even the build dependencies are included (does `--no-isolate` preclude use of offline builds?)
 - we can make CI faster, by installing the build deps from wheels, rather than building from source (as you point out in https://github.com/freedomofpress/securedrop-debian-packaging/issues/218#issuecomment-810386073)
 
Are there more benefits you'd like to call the team's attention to? 

Regarding the failing CI, I haven't actually taken a look at that part yet, although @zenmonkeykstop has some pretty solid leads above. Thanks for preparing, @kushaldas, looking forward to a deeper review.

kushaldas:
@conorsch 
>Why not store the new wheels for build and its dependencies inside the localwheels/ directory? We already have a directory for wheels we've built ourselves, but we're not using it in this PR. Conditional if/else logic has been added to most scripts, to support targeting either bootstrap/ or localwheels/, but it seems they'd work just as well unchanged, if the wheels were added to localwheels/.

Because I am signing the wheels for the bootstrap process, and I did not want to rebuild and resign the `sha256sums.txt`
file in the same PR. We can later move all wheels in the same place, but at that time, we should regenerate all the wheels, and sign the list. Also, I was not sure if there were any common wheel between localwheels and bootstrapping.
Keeping them separate makes things easier to create the process.

>What is build-requirements.txt used for? It was added in this PR, but I don't see it referenced in any of the test plans, or in the scripts. As of this PR, this repo now has three (3) Python requirements files. Coupled with the detached sig logic re-confirming the same SHA256 checksums as in the various component repos' build-requirements files, that's a lot of machinery to maintain. If we can remove build-requirements.txt, then we're back down to two (2), same as on the main branch.

We are using the `build-requirements.txt` file in the tests. This file contains the sha256sums of all the wheels required for the bootstrapping process. We have it mentioned in the `README` too in the steps. From now on when
we generate/rebuild our wheels, we will have to install the `build` tool in a virtualenv using this `build-requirements.txt` file. Just like we are using the similar file in Debian packaging, instead here we are using it for bootstrapping.

>Based on prior discussion, mostly in #218, the benefits of these changes appear to be:
>    we can ensure that the localwheels/*.whl files are built reproducibly, even when pip upgrades introduce breaking changes
>   we can potentially build the packages offline, given that even the build dependencies are included (does --no-isolate preclude use of offline builds?)

We need to use `--no-isolate` to make sure that the packages we are building get thier proper build dependencies. By default the tool [isolates](https://pypa-build.readthedocs.io/en/latest/#-no-isolation) the build environment.

>    we can make CI faster, by installing the build deps from wheels, rather than building from source (as you point out in #218 (comment))
>Are there more benefits you'd like to call the team's attention to?

As [mentioned](https://github.com/freedomofpress/securedrop-debian-packaging/issues/218#issuecomment-810421220) by upstream `pip` author, there will be no code execution during the installation from the wheels.

@zenmonkeykstop 
> with `git-lfs` installed first, most dependencies match, except for Cython:
> 
> ```
> ...
> Processing ./bootstrap/zipp-3.4.1-py3-none-any.whl
> Processing ./bootstrap/pip-21.0.1-py3-none-any.whl
> Processing ./bootstrap/setuptools-54.0.0-py3-none-any.whl
> ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
>     cython==0.29.22 from file:///home/circleci/project/bootstrap/Cython-0.29.22.tar.gz (from -r build-requirements.txt (line 2)):
>         Expected sha256 8c9eec7e9de2a30861ca347d0a149cc1482de12fc765fa06c414930e8ce20d0a
>              Got        df6b83c7a6d1d967ea89a2903e4a931377634a297459652e4551734c48195406
> ```
> 
> But! I think this is legitimate, the expected hash is for the .whl and the returned hash is actually correct for the tarball on a local checkout.

This was happening as we are using Ubuntu Focal with Python3.8 on CI instead of the Debian Buster with Python3.7. For now I added a Cython wheel for the Python3.8 (built via same steps on a Focal box). This unblocks us for the CI.

conorsch:
Made some small adjustments to the CI logic:

* No more Ubuntu VMs, uses Debian Buster containers throughout
* Removed py3.8 cython wheel, which was added to satisfy Ubuntu reqs
* Added pip-tools to the bootstrapping reqs file, so tarball hash is verified
* Removes tarball git-lfs installation, since Debian Buster has it in the repos

Please have a look, @kushaldas. If you agree we're in good shape, then please go ahead and merge. The next steps as I see them are:

1. Rebuild all wheels for all deps. Based on local testing, this will only modify ~3 deps, those with .so files shipped in the wheels. Builds are not currently blocked, so this isn't critical, but we can follow up shortly after merging to keep things consistent.
2. Consolidate the `bootstrap/*.whl` files into `localwheels/*.whl`. Let's wait and observe what the maintenance burden is with the setup as presented here—we may decide to keep it how it is.

Is there anything else you can point to, that we should handle shortly after merging? 
2.

conorsch (approved):
Approving. Requesting a visual spot-check from @kushaldas, since I pushed a few changes on top.

kushaldas:
@conorsch I followed the steps, and everything looks okay for the bootstrapping part of #218. Thank you for the review and the fixes :)

Issue #231, opened 2021-03-17
Reproducible builds: build_path showing variation


conorsch:
Discovered while debugging #228. See the diff from two package builds of `securedrop-client`, built from the 0.4.1 tag, but built in slightly different locations:

```
│ │ ├── ./opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client-0.4.1.dist-info/direct_url.json
│ │ │ │   --- /tmp/diffoscope_ee2yspmx/tmpiinoiy6_/0/1050.json
│ │ │ ├── +++ /tmp/diffoscope_ee2yspmx/tmpfc5ilrvz/0/1050.json
│ │ │ │┄ Similarity: 0.0%
│ │ │ │┄ Differences: {
│ │ │ │┄   "replace": {
│ │ │ │┄     "dir_info": {},
│ │ │ │┄     "url": "file:///home/user/src/securedrop-debian-packaging/build/debbuild/packaging/securedrop-client"
│ │ │ │┄   }
│ │ │ │┄ }
│ │ │ │ @@ -1,4 +1,4 @@
│ │ │ │  {
│ │ │ │      "dir_info": {},
│ │ │ │ -    "url": "file:///home/user/securedrop-debian-packaging/build/debbuild/packaging/securedrop-client"
│ │ │ │ +    "url": "file:///home/user/src/securedrop-debian-packaging/build/debbuild/packaging/securedrop-client"
│ │ │ │  }
```

Depending on the buildroot for the package build, we're seeing artifacts within the deb package change. That's not good. A few questions:

1. Can we catch this type of variations by enabling `+build_path` for reprotest in CI? 
2. Have we always had this problem, or did it start occurring recently? (We believe it's new.)
3. Are there documented or unexplained differences in tooling at work here? For example, are newer versions of pip and/or setuptools being pulled in dynamically? Does the behavior change inside a virtualenv vs using dist-packages?

Let's do some research and see what's what. 

<details>
<summary>full diffoscope output
</summary>

```
--- debugging/conor/securedrop-client_0.4.1+buster_all.deb
+++ debugging/mickael/securedrop-client_0.4.1+buster_all.deb
├── file list
│ @@ -1,3 +1,3 @@
│  -rw-r--r--   0        0        0        4 2021-03-17 18:20:12.000000 debian-binary
│ --rw-r--r--   0        0        0    33320 2021-03-17 18:20:12.000000 control.tar.xz
[...]

PR #243, opened 2021-04-07
Tests wheel reproducibility against version control

### LINKED ISSUE #241: CI should verify a new wheel in localwheels by rebuilding
When a maintainer will update the `localwheels` directory with newer version of wheels or source tarballs, the CI should verify those by rebuilding them.

## Why do we need this?

This CI step will verify that the sha256sum of any new wheel is exactly the same if the CI rebuilds it. Right now we have checks to verify if we generate wheels twice, both are same or not. But, we are missing the check to verify that the wheel sha256sum is the same of what the maintainer pushed manually in the PR.

conorsch:
We were already checking that building a wheel twice yields the same
checksum both times, but we weren't confirming that the checksum matched
what's already committed to version control. Let's do that.

Still leaving the 'reprotest' check enabled because it supports multiple
methods for checking reproducible, which should help us shake out
problems like #231.

Closes #241. 

### Testing

1. CI must be passing
2. Run `pytest -k test_wheel_builds_match_version_control tests` locally and confirm passing
3. Run `echo this-is-not-a-wheel > localwheels/MarkupSafe-1.1.1-cp37-cp37m-linux_x86_64.whl && gc -m 'TEMPORARY: broken wheel' localwheels/`, then re-run step 2, and confirm the test fails. 

Step 3 ensures that if a wheel build emits a different checksum from what's already version controlled, CI will fail on it.

### DIFF
--- a/tests/test_reproducible_wheels.py
+++ b/tests/test_reproducible_wheels.py
@@ -17,6 +17,7 @@ def get_repo_root():
     return top_level
 
 
+@pytest.mark.skip(reason="Comparing to version control is sufficient")
 @pytest.mark.parametrize("repo_name", REPOS_WITH_WHEELS)
 def test_wheel_builds_are_reproducible(repo_name):
     """
@@ -52,5 +53,8 @@ def test_wheel_builds_match_version_control(repo_name):
     repo_url = f"https://github.com/freedomofpress/{repo_name}"
     build_cmd = f"./scripts/build-sync-wheels -p {repo_url} --clobber".split()
     subprocess.check_call(build_cmd)
-    check_cmd = "git diff --exit-code".split()
-    subprocess.check_call(check_cmd)
+    # Check for modified files (won't catch new, untracked files)
+    subprocess.check_call("git diff --exit-code".split())
+    # Check for new, untracked files. Test will fail if working copy is dirty
+    # in any way, so mostly useful in CI.
+    assert subprocess.check_output("git status --porcelain".split()) == b""
--- a/tests/test_reproducible_wheels.py
+++ b/tests/test_reproducible_wheels.py
@@ -45,3 +45,12 @@ def test_wheel_builds_are_reproducible(repo_name):
     ]
     repo_root = get_repo_root()
     subprocess.check_call(cmd, env=cmd_env, cwd=repo_root)
+
+
+@pytest.mark.parametrize("repo_name", REPOS_WITH_WHEELS)
+def test_wheel_builds_match_version_control(repo_name):
+    repo_url = f"https://github.com/freedomofpress/{repo_name}"
+    build_cmd = f"./scripts/build-sync-wheels -p {repo_url} --clobber".split()
+    subprocess.check_call(build_cmd)
+    check_cmd = "git diff --exit-code".split()
+    subprocess.check_call(check_cmd)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -226,7 +226,7 @@ jobs:
           command: |
             make install-deps
             make lint-desktop-files
-            virtualenv -p python3 .venv
+            virtualenv -p /usr/bin/python3 .venv
             source .venv/bin/activate
             pip install -r test-requirements.txt
             sudo sed -i -re "292s/^(\s+).*\$/\1return _.prepend_to_build_command_raw('')/" /usr/lib/python3/dist-packages/reprotest/build.py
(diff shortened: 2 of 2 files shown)


conorsch:
CI is not passing: https://circleci.com/gh/freedomofpress/securedrop-debian-packaging/8074 likely due to #231. Marking as draft until resolved.

creviera:
yeah, https://github.com/freedomofpress/securedrop-debian-packaging/issues/231 is acting sus, however, i can't repro. I get the same reproducible wheel on different machines with different build roots :/

conorsch:
Shelving this work for now, to focus on #231. Have a draft PR open for comment. Assuming we can resolve #231, we should be unblocked for further work here. To be clear, I haven't actually verified that #231 (i.e. build_path variation) is what's breaking CI here, it's just a hunch.

conorsch:
Will take a closer look at CI now that https://github.com/freedomofpress/securedrop-debian-packaging/pull/246 is merged. At a glance, the CI run times are a good 10m longer or so now. Perhaps we should skip the old `reprotest` test for wheels and only use this new test, which effectively verifies reproducibility by comparing to vc.

kushaldas:
While trying to identify the differences between the wheels generated via the CI and our wheels,
I noticed the following diff.

```
│ │   The Directory Table (offset 0x1b):
│ │    1        src/markupsafe
│ │ -  2        /usr/include/x86_64-linux-gnu/bits
│ │ -  3        /usr/lib/gcc/x86_64-linux-gnu/8/include
│ │ +  2        /usr/lib/gcc/x86_64-linux-gnu/8/include
│ │ +  3        /usr/include/x86_64-linux-gnu/bits
│ │    4        /usr/include/x86_64-linux-gnu/bits/types
│ │    5        /usr/include
│ │ -  6        /usr/include/python3.7m
│ │ +  6        /usr/local/include/python3.7m
│ │    7        /usr/include/x86_64-linux-gnu/sys
```

If I look at the Dockerfile for `circleci/python:3.7-buster`, I can see it is based on [the official
Python image from Docker](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/python/images/3.7.10-buster/Dockerfile#L5)

Now that [image shows](https://github.com/docker-library/python/blob/master/Dockerfile-debian.template) that it installs Python executable build from source via the https://python.org
and uses the same as default Python.


But, all of our wheels are bulit via the system python provided by Debian Buster.

I now added another commit, where we are using the Debian provided Python.

tests/test_reproducible_wheels.py
```diff
@@ -45,3 +45,12 @@ def test_wheel_builds_are_reproducible(repo_name):
...
     repo_root = get_repo_root()
     subprocess.check_call(cmd, env=cmd_env, cwd=repo_root)
+
+
+@pytest.mark.parametrize("repo_name", REPOS_WITH_WHEELS)
+def test_wheel_builds_match_version_control(repo_name):
+    repo_url = f"https://github.com/freedomofpress/{repo_name}"
+    build_cmd = f"./scripts/build-sync-wheels -p {repo_url} --clobber".split()
+    subprocess.check_call(build_cmd)
+    check_cmd = "git diff --exit-code".split()
```
kushaldas, line 55: This is simple.

kushaldas (commented):
>At a glance, the CI run times are a good 10m longer or so now. Perhaps we should skip the old reprotest test for wheels and only use this new test, which effectively verifies reproducibility by comparing to vc.

Yes, I think removing the old reprowheels test is okay. We are verifying against the `lfs` checked in version here in this new code. 

@conorsch after you mark this PR as ready for review, we can do the formal review and merge this.

conorsch:
Ready for final review. See the most recent commits, @kushaldas, which disable the `reprotest` invocations but preserve the new logic of comparing the built wheels to version control. It's a bit aggressive to fail the tests if the git repo is dirty, since that'll often be the case for developers, but in the CI context that's precisely what we want: any deviation from our vc should raise an error.

kushaldas (approved):
All good. Approved. 

In future we may want to have a night repro test for the wheels.

tests/test_reproducible_wheels.py
```diff
@@ -52,5 +53,8 @@ def test_wheel_builds_match_version_control(repo_name):
     repo_url = f"https://github.com/freedomofpress/{repo_name}"
     build_cmd = f"./scripts/build-sync-wheels -p {repo_url} --clobber".split()
     subprocess.check_call(build_cmd)
-    check_cmd = "git diff --exit-code".split()
-    subprocess.check_call(check_cmd)
+    # Check for modified files (won't catch new, untracked files)
+    subprocess.check_call("git diff --exit-code".split())
+    # Check for new, untracked files. Test will fail if working copy is dirty
+    # in any way, so mostly useful in CI.
+    assert subprocess.check_output("git status --porcelain".split()) == b""
```
kushaldas, line 60: I should keep this command in mind. Thank you @conorsch.

PR #255, opened 2021-07-28
Bumps workstation kernel to 4.14.240

conorsch:
Includes a u2mfn version update, as well, which is required until Qubes
4.1 is final, when we can drop the u2mfn dependency entirely.


zenmonkeykstop (approved):
LGTM

PR #259, opened 2021-08-04
Bumps workstation kernel to 4.14.241

conorsch:
The previous version, based on 4.14.240, did not work well in Qubes VMs.
After adjusting the config, rebuilt on an incidentally slightly newer
kernel version, and that artifact behaves well in VMs. Removing the 240
version from the changelog, since it was never shipped to prod.

Also updates the control file fields, which update was omitted in #255.


creviera (approved):
lgtm

PR #261, opened 2021-08-12
Update signing key fingerprint

eloquence:
See key rotation blog post for more information:
https://securedrop.org/news/why-we-are-rotating-the-securedrop-release-key/

### DIFF
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -66,7 +66,7 @@ function verify_git_tag() {
     local t
     d="$1"
     t="$2"
-    prod_fingerprint="22245C81E3BAEB4138B36061310F561200F4AD77"
+    prod_fingerprint="2359E6538C0613E652955E6C188EDD3B7B22E6A3"
     if ! git -C "$build_dir" tag --verify "$PKG_VERSION" 2>&1 \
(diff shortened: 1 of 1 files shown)


conorsch:
Nice catch! However, it's a bit too soon to change to this setting by default: all the existing prod tags for the SDW components are still signed with the old key. For example:

```
$ make securedrop-export
WHEELS_DIR="/home/user/securedrop-debian-packaging/localwheels/" PKG_NAME="securedrop-export" ./scripts/build-debianpackage
[..snip..]
Failed to verify 0.2.6, not signed with 2359E6538C0613E652955E6C188EDD3B7B22E6A3
make: *** [Makefile:29: securedrop-export] Error 2
```

So, much as I hate to say it, I think the solution for now is to permit _either_ signature as valid. Ultimately, I'd prefer that we switch to using a keyring-based approach to validate the tags, so any current maintainer's sigs are valid, but that's a separate discussion.

conorsch:
Ah, nevermind: the expired sig on the current tags means that even on the `main` branch, the logic fails, because `git tag --verify` returns non-zero if the key is expired. So, fine to merge as-is.

conorsch (approved):

PR #262, opened 2021-08-16
Update maintainer pubkeys

conorsch:
Removes @rmol & @emkll, adds @zenmonkeykstop.
These pubkeys are authorized to sign the wheel hashes,
rather than re-using the release key.


creviera (approved):
lgtm

PR #264, opened 2021-08-27
update idna from 2018 release to current

creviera:
# Description

Adds wheels for current idna prod dependency. This is not in response to a CVE but rather regualar, periodic maintenance of outdated packages. Since we plan to release a new version of the SDW soon with other dependency updates, we might as well pull this in too while doing QA for the upcoming release.

If there isn't time for this, we can wait until after the major SDW release.

# Todo

- [x] Add diff review for idna https://github.com/freedomofpress/securedrop-debian-packaging/wiki/idna-2.8-to-3.2


kushaldas (approved):
Added the diff-review, all looks good.

PR #268, opened 2021-09-01
Docs for how to skip signature verification

creviera:
# Description

Add docs for how to build a debian package for development and test purposes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,50 @@ gpg --verify repo/public/dists/buster/Release.gpg repo/public/dists/buster/Relea
 
 Once the PR is merged, the new packages will be picked up by a script and deployed to https://apt.freedom.press within 30 minutes.
 
+## Build a package for development or test (skip signature verfication)
+
+Checkout the project you intend to package and enter that directory:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -237,6 +237,50 @@ gpg --verify repo/public/dists/buster/Release.gpg repo/public/dists/buster/Relea
...
+cd securedrop-foobar
+```
+
+If there is a dev or rc tag you are testing, then checkout that tag, e.g.
+
+```
+git checkout x.y.zrc1
+```
+
+Update `setup.py` with the rc or dev version number, e.g. `0.5.0dev1` or `0.5.0rc1`. In order for the debian packaging tool to work, you cannot use a `-` or `.` or `_` between the vesion number and the `dev` or `rc` portion of the string. Now you can generate a tarball to be used in the build process:
```
kushaldas, line 255: This is super important to remember for us.

kushaldas (approved):
This is a good update and will be very helpful in future for any new developer and even for us.

Thank you.

PR #269, opened 2021-09-02
update paths for build reqs for various projects

creviera:
# Description

Just like we did here: https://github.com/freedomofpress/securedrop-debian-packaging/pull/266 use the latest path for the build requirements file

### DIFF
--- a/securedrop-export/debian/rules
+++ b/securedrop-export/debian/rules
@@ -3,7 +3,10 @@
 %:
 	dh $@ --with python-virtualenv
 
+REQUIREMENTS_FILE=requirements/build-requirements.txt
+
 override_dh_virtualenv:
+	test -e $(REQUIREMENTS_FILE)
 	dh_virtualenv \
 		--python /usr/bin/python3 \
 		--setuptools \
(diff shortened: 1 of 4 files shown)


securedrop-export/debian/rules
```diff
@@ -3,7 +3,10 @@
 %:
 	dh $@ --with python-virtualenv
 
+REQUIREMENTS_FILE=requirements/build-requirements.txt
+
 override_dh_virtualenv:
+	test -e $(REQUIREMENTS_FILE)
```
kushaldas, line 9: This will fail (you can see that in https://app.circleci.com/pipelines/github/freedomofpress/securedrop-debian-packaging/1259/workflows/525b72d5-646c-486a-88b4-1c150e3e90b7/jobs/9838/steps ) because at this step, Debian packaging system is not in a directory with the source files extracted.  It took me some time to understand why it was failing and passing for `securedrop-log`. 

General rule for keeping packaging easy is to not modify/hack these steps heavily.

conorsch: Based on local testing for #266, I'm confident that an existence check at this point in the build will pass/fail. The fact that it's failing in CI now for `securedrop-export`, but not for `securedrop-log`, which was also updated here, is puzzling to me. Locally, builds for both pass based on this branch. I'll investigate more closely.

conorsch: The problem was that `securedrop-export` was building from the most recent signed tag on PRs into this repo, so the tarball didn't include the reqs file changes. Thus the mismatch, and the error. I've pushed a commit to address, CI should be passing now.

conorsch:
I also tacked on a commit bringing `securedrop-client` up to spec. That should be all of them now.

conorsch (approved):
Thanks for following up on #266 , to get things straightened out across the board. Wish we didn't have to do this, but for resiliency's sake it's definitely worthwhile.

PR #280, opened 2021-11-24
upgrade pip to version 21.1 or later [CVE-2021-3572]

creviera:
# Description

Upgrades our bootstrap dependency `pip` from 21.0.1 to 21.3.1 to address [CVE-2021-3572:](https://github.com/advisories/GHSA-5xp3-jfq3-5q8x):

> A flaw was found in python-pip in the way it handled Unicode separators in git references. A remote attacker could possibly use this issue to install a different revision on a repository. The highest threat from this vulnerability is to data integrity. This is fixed in python-pip version 21.1.

# Test Plan

- [ ] Diff review pip 21.0.1 to 21.3.1
- [ ] CI passes


gonzalo-bulnes:
I'll pick this diff review. :hand:

gonzalo-bulnes (commented):
**Instructions**

I'm following the step-by-step defined in https://github.com/freedomofpress/securedrop/wiki/Dependency-specification-and-update-policies#step-by-step-procedure

**Findings**

@creviera I list them as a checklist so that you can go through them and check they're okay.

- [ ] :skull_and_crossbones: Someone found funny to play with [Trojan Source][ts] - GitHub's tooling highlighted it straight away :sunglasses: ([code with Unicode bidirectional control characters][ts-pip]) The less nice part is that it may start triggering automated checks unnecessarily.

----

Ok... this is going to take longer than expected. A lot of line endings were changed, and linting was performed across the project.

:warning: I fell back to using GitHub to highlight the diffs when I noticed that.

I've set up a "diff visualisation" in [this code review fork/PR](https://github.com/gonzalo-bulnes/diff-review-pip/pull/1), which contains all the files that have changed *excluding* the entire `docs/` directory.

GitHub's PR review tooling does consider line endings as changes, so most files' diffs are not very helpful. **However** this [other comparison](https://github.com/pypa/pip/compare/21.0.1...21.3.1) doesn't suffer from that inconvenient, so I started switching back and forth between the two. (Reading a file in the latter, then marking it as viewed in the former.)

I'll stop here for today, and touch base tomorrow, I suspect this is not completely unheard of as a situation and there may be workarounds I'm not thinking of. @creviera 

  [ts]: https://trojansource.codes
  [ts-pip]: https://github.com/pypa/pip/commit/ea761a6575f37b90cf89035ee8be3808cf872184#diff-b897336dda0579c40f3623b68eb6652ded79e206f4cf1af7ceb24b683fd46ff1R615

creviera:
@gonzalo-bulnes and I discussed today that we should just update to `21.1` instead of `21.3.1` for a smaller diff review so we can  resolve this vulnerability. We can look into prioritizing future diffs to get us closer to the latest when there's less of a time constraint.

creviera:
Based on the work @gonzalo-bulnes has already done here and the outcome of our team discussion, this PR is ready for final review/ approval.

conorsch:
Thanks, both. Joining the conversation late here: since this dependency is `pip`, which is used all over the Python ecosystem, we don't need a full diff review of the changes. We _do_ want to pin a specific hash, though, so the changes as written appear to be good. 

Our policy on what to diff review doesn't mention the exemption I'm pointing it—but I'll update the docs in the SD wiki on dependency updates and link back here when done.

conorsch (approved):
Approving to unblock ongoing release mechanics. I still owe docs updates related to this change, and will follow up with those shortly.

conorsch:
Added a note to https://github.com/freedomofpress/securedrop/wiki/Dependency-specification-and-update-policies

> (If the dependency is `pip`, you can skip this step.)

I took a quick look at https://github.com/freedomofpress/securedrop-debian-packaging/blob/28e6e17ed3451ba37223754e0d5394231feff6a3/bootstrap-sha256sums.txt and I don't think we can blanket-allow-list all those deps. For something as fundamental as pip, it's reasonable to trust PyPI, as long as we continue to pin checksums.

PR #287, opened 2022-02-02
Have CI automatically update Tor packages in apt-test

legoktm:
## Status

Ready for review

## Description

Instead of maintaining and running a custom molecule playbook to pull
down Tor packages, we can use reprepro's built-in update functionality.

Once <https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/128>
is merged, we can run `reprepro update` to automatically check the
upstream Tor apt repo, pulling down new debs if they're available.

The CI job checks out the securedrop-dev-packages-lfs repository, runs
`reprepro update` and copies over the newly fetched debs. If there's a
difference, it'll commit the new packages and push.

In the future we could publish an image with the reprepro, etc. packages
pre-installed so we don't have to do it at runtime.

Refs <https://github.com/freedomofpress/securedrop/issues/6233>.

## Test plan
* Step through the commands manually inside a container to make sure they work (no obvious typos, etc.)
* Besides that I'm not really sure... could merge this, wait for a new Tor upstream release, and then see if it works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -262,6 +262,36 @@ jobs:
             sudo sed -i -re "292s/^(\s+).*\$/\1return _.prepend_to_build_command_raw('')/" /usr/lib/python3/dist-packages/reprotest/build.py
             pytest -vvs tests/test_reproducible_debian_packages.py
 
+  reprepro-update-tor:
+    docker:
+      - image: debian:bullseye
+    steps:
+      - *addsshkeys
+      - run:
+          name: clone and run reprepro update
+          command: |
+            # TODO: Publish image with these packages pre-installed
+            apt-get update
+            apt-get install -y reprepro ca-certificates dctrl-tools git git-lfs openssh-client
+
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -262,6 +262,35 @@ jobs:
...
+    steps:
+      - run:
+          name: clone and run reprepro update
+          command: |
+            # TODO: Publish image with these packages pre-installed
+            apt-get update
+            apt-get install -y reprepro ca-certificates dctrl-tools git git-lfs openssh-client
+
+            # Clone the dev repo and configure it
+            git clone git@github.com:freedomofpress/securedrop-dev-packages-lfs.git
```
legoktm, line 277: I'm assuming that this will just work and Circle CI already provisions access to the necessary credentials? I didn't see anything explicit in the workstation nightly jobs...

zenmonkeykstop:
RSS feed seems like a good option for a first pass!

PR #297, opened 2022-04-06
Add @rocodes pubkey

rocodes:
Adds my pubkey to packaging repo (can be verified at https://freedom.press/people/rowen-s)


creviera:
lgtm, but just realized my key is expired, so I'll append an update commit to this pr

conorsch (approved):
Both pubkeys LGTM, so I'm approving. The reprotest suite is failing on `securedrop-log`, which is confounding to me, but clearly unrelated to these changes. I'll file a follow-up to investigate the test failure.

creviera:
Thanks Conor. Going to push forward with this and we can investigate further after making some headway with SDW relesases.

Issue #298, opened 2022-04-06
Reprotest is failing for securedrop-log


conorsch:
First reported during review of https://github.com/freedomofpress/securedrop-debian-packaging/pull/297 ; example test failure is here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-debian-packaging/1531/workflows/928739e8-0fca-4ada-931a-218e17359b51/jobs/11788

Unclear why just `securedrop-log` is failing the reprotest checks now.

legoktm:
The failure output is:
```
New python executable in /tmp/reprotest.hILQ39/const_build_path/build/debbuild/packaging/securedrop-log/debian/securedrop-log/opt/venvs/securedrop-log/bin/python3
Also creating executable in /tmp/reprotest.hILQ39/const_build_path/build/debbuild/packaging/securedrop-log/debian/securedrop-log/opt/venvs/securedrop-log/bin/python
Installing setuptools, pip, wheel...done.
Looking in links: /tmp/reprotest.hILQ39/const_build_path/localwheels/
Processing /tmp/reprotest.hILQ39/const_build_path/localwheels/redis-3.3.11-py2.py3-none-any.whl
Installing collected packages: redis
Successfully installed redis-3.3.11
Looking in links: /tmp/reprotest.hILQ39/const_build_path/localwheels/
Processing /tmp/reprotest.hILQ39/const_build_path/build/debbuild/packaging/securedrop-log
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  Preparing metadata (setup.py) ... error
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
Traceback (most recent call last):
  File "/usr/bin/dh_virtualenv", line 110, in <module>
    sys.exit(main() or 0)
  File "/usr/bin/dh_virtualenv", line 93, in main
    deploy.install_package()
  File "/usr/lib/python2.7/dist-packages/dh_virtualenv/deployment.py", line 260, in install_package
    subprocess.check_call(self.pip(package), cwd=os.path.abspath(self.sourcedirectory))
  File "/usr/lib/python2.7/subprocess.py", line 190, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #307, opened 2022-04-20
Pass --no-clean-on-error to reprotest for easier debugging

legoktm:
When reprotest fails, leave the environment behind for easier
interactive debugging of specific commands that failed. In CI everything
is thrown away anyways, so this should only affect people locally
running reprotest.

Refs #298.


creviera (approved):
lgtm

PR #309, opened 2022-04-22
Have CI delete nightlies older than 2 weeks

legoktm:
Have CI run a quick Python script that will delete workstation nightlies
after 2 weeks to keep the repository size down.

This job requires the previous nightly builds to succeed, so if the
nightlies break for whatever reason, deletion will also be paused.

## Test plan
* [ ] Check out this repo and securedrop-dev-packages-lfs side by side
* [ ] Manually run the commands CI would, up till the point where the commit is created (but don't push)


creviera (approved):
## Test plan
* Check out this repo and securedrop-dev-packages-lfs side by side
* [x] Manually run the commands CI would, up till the point where the commit is created (but don't push)

Issue #312, opened 2022-05-02
Release `securedrop-workstation-config` 0.1.7


eaon:
The current `securedrop-workstation` tests expect the `sd-viewer` VM to have the most recent version of `securedrop-workstation-config/mimeapps.list.sd-viewer` in place, which it is not for 0.1.6.

This will pull in the following change:
* https://github.com/freedomofpress/securedrop-debian-packaging/commit/2dcb936f07773dfe643b1730d717a188728dd79c

## QA Test Plan (when package is on apt-test)

- [x] mp4 tests pass when you run `make test`
- [x] You can play an mp4

creviera:
For reference on the release process for metapackage components, see https://github.com/freedomofpress/securedrop-workstation#release-a-subproject

creviera:
Based on what we did for the [0.1.6 release](https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/99), it looks like we should follow this plan:
1. prod-sign a `0.2.23` tag for `securedrop-debian-packaging`
2. build `securedrop-workstation-config` and push to the dev lfs repo
3. follow test plan defined in the lfs PR (it will include playing an mp4 file and running `make test` -- we know this works already after testing nightlies)
4. publish the same package to the prod lfs repo
5. send out announcement about the fix

eaon:
Tests pass and mp4 audio file is playing in a DispVM started by the client!

eaon:
Resolved with https://github.com/freedomofpress/securedrop-debian-packages-lfs/pull/78 :tada:

PR #313, opened 2022-05-02
Add my pubkey to the repo

eaon:
Talked to @creviera about how to go about releasing `securedrop-workstation-config` (#312) and it came up that it would make sense for my pubkey to be in the repo (for the future anyway)


creviera (approved):
lgtm

Issue #308, opened 2022-04-22
Move debian/ directories into git repositories where rest of the code is


legoktm:
## Progress
* [ ] securedrop-client
* [ ] securedrop-export
* [ ] securedrop-log: https://github.com/freedomofpress/securedrop-log/pull/26, https://github.com/freedomofpress/securedrop-debian-packaging/pull/316
* [ ] securedrop-proxy

## Rationale
For the main Python projects (securedrop-client, securedrop-log, etc.), we are currently maintaining the debian/ directory separate from the main project. This is similar to the workflow used when you have a distinct upstream that releases software, and someone in Debian who then packages it and distributes it.

However in our case, there is no distinction between the "upstream" project and the packaging - they're done by the same people at the same time. We make our releases in the format of Debian packages - if there's a code change a new package will be built and if there's a packaging change, a new release will be issued! In Debian terminology this is called a "native" package.

This leads to extra work split across multiple repos, e.g. to release 0.7.0 of securedrop-client, it required two PRs, one to bump the version (https://github.com/freedomofpress/securedrop-client/pull/1483) and then another in this repo to bump the version in d/changelog (https://github.com/freedomofpress/securedrop-debian-packaging/commit/5c71d68f59bfe7b09f212bbbe3bdca12c45264f9). Or consider if you wanted to add a new file to be installed, you'd have to make your PR and then do another PR in this repo to add it to `d/<package>.install`. And we also don't test new changes against the packaging until the nightly runs, which is subpar.

My proposal is that we just move the debian/ directory for these projects into the corresponding Git repos (WIP example: https://github.com/freedomofpress/securedrop-log/commits/debian-dir). For now, this repo will still be the main entrypoint for building packages. The immediate benefit will be that making releases and packaging changes doesn't require two coordinated PRs. In the build logic, we can skip everything about building a sdist tarball and then repacking it, we'll just use the git repo tree as our build directory (see my WIP: https://github.com/freedomofpress/securedrop-debian-packaging/commit/1f9c4c2942a020cd43a2531c184f7f1a35edc5ed).

This change also unlocks more possible future steps. The main thing is that each git repo now mostly (minus some verification stuff and localwheels) contains everything needed for packaging, which would allow us to use git-buildpackage workflows. We could also run CI for building a package on every commit/PR, and rather than doing nightlies (or we keep nightlies too? idk), just build new packages after every commit pushed to main.

conorsch:
[...]

PR #315, opened 2022-05-06
Build native packages with debian/ in source tree

legoktm:
The end goal is that each repository has all the necessary files to
build a Debian package, without needing this repository. This gets us
partially there for metadata in the debian/ directory.

Currently we build a sdist tarball and use it as the source tree for
packaging. This has some nice properties in that we can exclude some
unnecessary files and lets us control timestamps for reproducibility.
This comes at a higher cognitive cost of needing to know everything that
ends up in a sdist tarball and complexity to reset timestamps.

With this change, we now use the Git repo tree as the source tree for
packaging. Steps like copying over a debian/ directory from this
repository and installing a platform-specific changelog are guarded on
files existing so they can be removed/adjusted on a repo-by-repo basis.

Refs #308.

## Test plan
* [x] CI passes
* [ ] Manually build a package or two with `make [package]`
* [ ] Manually build a package from a git checkout with `PKG_PATH=../securedrop-whatever make [package]`
* [ ] Manually build securedrop-log by checking out the the `debian-dir` branch and `debian-remove-sdlog` branch in this repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,14 +58,6 @@ common-steps:
         echo $VERSION_TO_BUILD > ~/packaging/sd_version
         echo 'export VERSION_TO_BUILD=$(cat ~/packaging/sd_version)' >> $BASH_ENV
 
-  - &makesourcetarball
-    run:
-      name: Create source tarball
-      command: |
-        cd ~/packaging/securedrop-*
-        git checkout $VERSION_TO_BUILD
-        python3 setup.py sdist
-
   - &clonesecuredropclient
     run:
       name: Clone the repository to be packaged
@@ -135,7 +127,7 @@ common-steps:
     run:
       name: Build debian package
       command: |
-        export PKG_PATH=~/packaging/$PKG_NAME/dist/$PKG_NAME-$VERSION_TO_BUILD.tar.gz
+        export PKG_PATH=~/packaging/$PKG_NAME/
         export PKG_VERSION=$VERSION_TO_BUILD
         make $PKG_NAME
         ls ~/project/build/debbuild/packaging/*.deb
@@ -333,7 +325,6 @@ jobs:
       - *installdeps
       - *clonesecuredroplog
       - *getnightlyversion
-      - *makesourcetarball
       - *builddebianpackage
 
   build-nightly-buster-securedrop-log:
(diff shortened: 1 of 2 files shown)


creviera (changes requested):
How can we test a dev build with this change? Normally, we would do the following (using `securedrop-client 0.7.1dev0` for example):
1. in the project repo, run `./update_version.sh 0.7.1dev0`
2. run `python3 setup.py sdist` to build a 0.7.1dev0 tarball
3. in the packaging repo, run `PKG_PATH=../securedrop-client/dist/securedrop-client-0.7.1.dev0.tar.gz make securedrop-client`

:heavy_check_mark: build is successful on `main`
:x: build is successful on `debian-in-tree`

legoktm:
The idea is that we will no longer create the sdist tarball. So in your example the workflow is now:

1. in the project repo, run `./update_version.sh 0.7.1dev0`
2. in the packaging repo, run `PKG_PATH=../securedrop-client/ make securedrop-client`

The updated commit I just pushed will make it so if PKG_PATH points to a tarball, it'll continue to work for backwards-compatibility reasons, so your original workflow should work now (mainly because it's still in various CI pipelines).

You can also see how it functions with a debian/ folder in the repository itself with `PKG_GITREF=debian-dir make securedrop-log` vs `PKG_GITREF=main make securedrop-log` (using debian/ from securedrop-debian-packaging).

creviera (approved):

creviera:
My bad, I had some changes in my local env, so here are my actual results:

> 1. in the project repo, run `./update_version.sh 0.7.1dev0`
> 2. in the packaging repo, run `PKG_PATH=../securedrop-client/ make securedrop-client`

I tried this for `securedrop_proxy` and it doesn't create the package with the dev string as expected, but works otherwise.

I'll create a followup issue for being able to do something like:
> 1. in the project repo, run `./update_version.sh 0.7.1dev0`
> 2. in the packaging repo, run `PKG_VERSION=1.2.3dev0 ./scripts/update-changelog`
> 2. in the packaging repo, run `PKG_PATH=../securedrop-client/ make securedrop-client`

> The updated commit I just pushed will make it so if PKG_PATH points to a tarball, it'll continue to work for backwards-compatibility reasons, so your original workflow should work now (mainly because it's still in various CI pipelines).

This works, so as long as there is _a_ way to build a dev package, this lgtm.

Let me know if you want to pair so we can see what's different between your tests and mine.

legoktm:
Once we move debian/ into the repositories themselves, we can teach update_version.sh how to increment debian/changelog as well: https://github.com/freedomofpress/securedrop-log/pull/26/commits/2fa9b6f73e6c57d5a68fc262743b2eb99632a6bf

PR #301, opened 2022-04-13
[SPIKE] Update workstation-grsec for Bullseye

conorsch:
Builds a new `securedrop-workstation-grsec` metapackage for
Debian 11 Bullseye, intended for use with Qubes 4.1. Major changes:

  * upgrades kernel series 4.14.x -> 5.15.x
  * removes u2mfn references (not necessary under Qubes 4.1)

This change was made indelicately to the build logic: no care was taken
to preserve functionality for the oldstable Buster grsec logic. As such, we should not merge as-is. The work is presented in draft form related to spike in https://github.com/freedomofpress/qubes-template-securedrop-workstation/pull/24

PR #317, opened 2022-05-10
Add a bullseye builder

legoktm:
Duplicate the current dockerfiles/ directory into a separate buster
and bullseye directories with the understanding that the buster
one is going to go away pretty soon, so it's not worth abstracting
the logic to be platform agnostic.

I dropped some packages and the timezone step that were already present
in the base `debian:bullseye` image.

I added my PGP key to the trusted set and rebuilt the Cython wheel for Python 3.9,
so now the bootstrap is installable in the bullseye container.

Note: I have not yet pushed the image to quay.io.


build-requirements.txt
```diff
@@ -1,6 +1,6 @@
 build==0.3.0 --hash=sha256:3fe8fac398ae76f534dee92c0db01c2960e271094f8cd52aa9d24a633c244e59
 click==7.1.2 --hash=sha256:3b0769bd65ca00ffcb57975e750fac079639a3d3ce393c5a0d4c65123141c87a
-cython==0.29.22 --hash=sha256:8c9eec7e9de2a30861ca347d0a149cc1482de12fc765fa06c414930e8ce20d0a
+cython==0.29.22 --hash=sha256:8c9eec7e9de2a30861ca347d0a149cc1482de12fc765fa06c414930e8ce20d0a --hash=sha256:df6b83c7a6d1d967ea89a2903e4a931377634a297459652e4551734c48195406
```
creviera, line 3: did this already get diff-reviewed?

creviera:
@legoktm you can add your key to https://github.com/freedomofpress/securedrop-debian-packaging/tree/main/pubkeys

build-requirements.txt
legoktm: It's the hash of the tarball that's already in the repository, so I assume it was reviewed.

creviera:
Does this conflict with https://github.com/freedomofpress/securedrop-debian-packaging/pull/301 in any way?

build-requirements.txt
creviera: ah, i see that now!

creviera (approved):
Spoke to @legoktm, did a code walkthrough, and agreed to do more testing (than just building the container locally) post merge so that we can confirm that we can successfully build a workstation component for bullseye. Since we are about to build 9 components, we'll have plenty of opportunity to work out any issues that may arise.

re: https://github.com/freedomofpress/securedrop-debian-packaging/pull/317#issuecomment-1124050120, this shouldn't conflict with https://github.com/freedomofpress/securedrop-debian-packaging/pull/317#issuecomment-1124050120, which adds bullseye build logic for the qubes workstation template.

legoktm:
quay.io/freedomofpress/packaging-debian-bullseye:2022_05_10 has now been pushed.

PR #320, opened 2022-05-16
Only install reprotest when needed

### LINKED ISSUE #318: Avoid installing 4GB of dependencies when running `make install-deps`
`make install-deps` pulls in mostly-bootstrapping tools, but it also brings in reprotest, which depends on diffoscope, which depends on like, the entire world with stuff like haskell, java, android tools, etc. being pulled in.

```
1 upgraded, 1102 newly installed, 0 to remove and 0 not upgraded.
Need to get 1189 MB of archives.
After this operation, 4728 MB of additional disk space will be used.
```

without reprotest we go down to:
```
Need to get 80.1 MB of archives.
After this operation, 222 MB of additional disk space will be used.
```

I propose that we only install reprotest in the two CI pipelines that need it and remove it from `install-deps`. It's possible that this was pulling in some build dependency that is actually needed but never explicitly specified - catching that would be a good thing overall.

legoktm:
This will dramatically shrink the number of dependencies needed when
running `install-deps` from 4GB to 200MB because all of the diffoscope
dependencies aren't needed.

We only need reprotest for three different CI pipelines, so manually install
it there instead of everywhere.

Fixes #318.

----
Without this change:
Screenshot 2022-05-16 at 15-42-57 Pipelines - freedomofpress_securedrop-debian-packaging

With this change:
Screenshot 2022-05-16 at 15-48-49 Pipelines - freedomofpress_securedrop-debian-packaging

So we're looking at a ~2min speed up for the main build jobs, which should be a noticable speedup on the various component repositories. And spinning up a new container locally will no longer have to download 4G worth of packages...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -217,6 +217,7 @@ jobs:
           name: install test requirements, run linters, and run tests
           command: |
             make install-deps
+            sudo apt-get install reprotest -y
             make lint-desktop-files
             virtualenv -p /usr/bin/python3 .venv
             source .venv/bin/activate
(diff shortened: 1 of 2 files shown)


creviera (approved):
<3 all that time savings! thx @legoktm!

Issue #321, opened 2022-05-16
Build bullseye packages


legoktm:
We need bullseye packages for the upcoming buster EOL. An initial builder was added in #317.

For each package:
* [ ] Configure CI to run tests on bullseye/Python 3.9 (in component repo)
* [ ] Build the package under bullseye:
  * [ ] Remove `--setuptools`, no-op in buster and bullseye
  * [ ] Rebuild localwheels for Python 3.9
  * [ ] Upgrade to debian-compat = 12 (see https://nthykier.wordpress.com/2019/01/04/debhelper-compat-12-is-now-released/)
* [ ] Configure CI to build bullseye nightlies (in packaging repo)
* [ ] Configure CI to build bullseye packages (in component repo)

legoktm:
With #322, I tested building a few packages and it seems we're ready to work on individual packages now. The plan is for me to pair with different people who expressed interest, and each pair can tackle 1-2 packages. I think our success criteria will be that the package was built successfully and looks reasonable (like, it contains the necessary files). Testing is going to be hard since we can't use debdiff/diffoscope since all the files moved from python3.7 -> python3.9, and we really need *all* the packages to effectively test SDW. So that'll have to be a collective step once we're done.

creviera:
Super helpful checklist @legoktm, thanks! Tomorrow, it would make sense to start with the `securedrop-keyring` package (https://github.com/freedomofpress/securedrop-debian-packaging/issues/319) if you don't mind pairing with someone on that (I volunteer for this first one if no one else does).

creviera:
At this point, we have all the nightlies building for bullseye, so closing this tracking issue. Please reopen if I missed anything.

PR #322, opened 2022-05-16
More bullseye fixes

legoktm:
First, explicitly install python3-virtualenv
    
The version of dh_virtualenv on bullseye allows using `python3-venv` and
if present, won't install `python3-virtualenv`. Since we still need
virtualenv, install it explicitly for now until we can transition to
using venv.

Second, automatically determine platform being built for
    
Instead of requiring `PKG_PLATFORM` to manually be set, we can get the
current platform from `/etc/os-release`.

The build script will also now adjust the platform in the version
number, so we don't need to maintain separate changelogs that are
identical except for version.

Refs #321.

## Test plan
* [ ] Spin up a bullseye container/VM, run `make install-deps`
* [ ] Edit securedrop-log/debian/rules to remove the `--setuptools` flag to dh_virtualenv
* [ ] run `PKG_GITREF=main make securedrop-log` and watch as you get a bullseye deb!
* [ ] Now switch back to a buster container/VM, run `make install-deps`
* [ ] Run `PKG_GITREF=main make securedrop-log` and you still get a buster deb!

### DIFF
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -27,8 +27,8 @@ TOP_BUILDDIR="$PWD/build/debbuild/packaging"
 mkdir -p "$TOP_BUILDDIR"
 rm -rf "${TOP_BUILDDIR:?}/${PKG_NAME}"
 mkdir -p "${TOP_BUILDDIR}/${PKG_NAME}"
-# Move changelog into place (we have separate changelogs for each platform)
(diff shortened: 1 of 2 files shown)


creviera:
I can review this and will test it by building `securedrop-keyring`, which we need to do anyway.

creviera (approved):
lgtm

Issue #319, opened 2022-05-16
Release new securedrop-keyring package


creviera:
Bump the public keyring in the `securedrop-keyring` workstation package for both Buster and Bullseye (the Bullseye keyring package will be used once we're ready for users to switch to Qubes 4.1). 

Note that there's a separate `securedrop-keyring` package for the servers.

creviera:
Closing now that the keyring package is on dev lfs. We can track the rest of the workstation components using: https://github.com/freedomofpress/securedrop-debian-packaging/issues/321

PR #323, opened 2022-05-17
Update signing pubkey

creviera:
# Description

Once this is merged, we can tag `securedrop-debian-packaging` and build a new `securedrop-keyring` debian package (2! one for buster, one for bullseye) an push to `securedrop-dev-pacakges-lfs`.

Towards https://github.com/freedomofpress/securedrop-debian-packaging/issues/319

# Test Plan

- [x] Changelog looks good
- [x] Confirm `release_key.gpg` (confirm that this is the 2021 key with new expiration date)
- [x] Confirm `securedrop-keyring.gpg` (confirm that the keyring only contains the 2021 key with the new expiration date)


legoktm:
AFAICT the file in `pubkeys/release_key.pub` is the correct key, but it has a different checksum than install_files/ansible-base/roles/install-fpf-repo/files/fpf-signing-key-2021.pub in the securedrop repo. Can we re-use that exact same file?

The gpg keyring file is the exact same, which made verification a breeze:
```bash
$ sha256sum securedrop-keyring/securedrop-keyring.gpg
746466799881fb2546a1cdbe89b07b6c65b6013f4885d1fe122a3f3da3728332  securedrop-keyring/securedrop-keyring.gpg
$ sha256sum install_files/securedrop-keyring/etc/apt/trusted.gpg.d/securedrop-keyring.gpg
746466799881fb2546a1cdbe89b07b6c65b6013f4885d1fe122a3f3da3728332  install_files/securedrop-keyring/etc/apt/trusted.gpg.d/securedrop-keyring.gpg

creviera:
To make it so you can skip the step of importing the key and checking that it's the correct one, and instead just check that the sha256sum matches the one added to the server, I'll copy it. (I would just import it and verify that everything looks correct anyway.)

legoktm (approved):
Yay, key, keyring, and changelog all look good! Just waiting on CI

PR #327, opened 2022-05-24
update changelog for securedrop-workstation-config bullseye 0.1.7

creviera:
Towards https://github.com/freedomofpress/securedrop-debian-packaging/issues/321

* Add bullseye changelog for securedrop-workstation-grsec
* Add job for building securedrop-workstation-grsec in CI

Cross link to https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/159

Once merged, you can push to `pleasebuildnightlies` to test if CI properly builds the package.


legoktm:
I don't think adding a new changelog is necessary, the package is going to be the same regardless of buster vs bullseye, so it can reuse the existing buster changelog. All that really needs to happen is adding to CI.

PR #332, opened 2022-05-26
Fix building packages when PKG_PATH doesn't end in PKG_NAME

legoktm:
There's a subtle bug in which we (read: I) incorrectly assumed that
PKG_PATH will always end in PKG_NAME, so copying it over into
$TOP_BUILDDIR will create the appropriately named subdirectory and
everything is happy.

This assumption is wrong in all the component repository CI pipelines,
where the path points to a directory named "project". Adjust the `cp`
command with the magic `-T` flag to stick the files in the correctly
named directory that we expect later on.

Thanks to Gonzalo for helping with debugging.


gonzalo-bulnes (approved):
The trailing slashes are for consistency and ease of reasoning. :bulb: Today I learned about `cp -T` :slightly_smiling_face:

Issue #333, opened 2022-05-26
Buster dev build of securedrop-client failing due to compat issue


eloquence:
## Steps to reproduce

On an up-to-date Buster VM, attempt to build a dev package following the [instructions in the README](https://github.com/freedomofpress/securedrop-debian-packaging#build-a-package-for-development-or-test-skip-signature-verfication) fails.

## Expected behavior

Package is built.

## Actual behavior

As of #329, build fails with error:

>  debian/rules build
> make[1]: Entering directory '/home/user/securedrop-debian-packaging/build/debbuild/packaging/securedrop-client'
> dh build --with python-virtualenv
>    dh_update_autotools_config
>    dh_autoreconf
>    dh_auto_configure
> dh_auto_configure: Please use the third-party "pybuild" build system instead of python-distutils
> dh_auto_configure: This feature was removed in compat 12.
> make[1]: *** [debian/rules:4: build] Error 2
> make[1]: Leaving directory '/home/user/securedrop-debian-packaging/build/debbuild/packaging/securedrop-client'
> dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2

## Analysis

Lowering the compat level in this file to 11 resolved the issue for me.

creviera:
I can confirm that we are no longer able to build a bullseye or buster package by pointing our builder at a source tarball via:
`PKG_PATH=path/to/source/tarball make package-name`.

I can also confirm that switching to compat 11 fixes this issue for both buster and bullseye. 

We could either lower the compat to 11 (making debhelper more backwards compatible) or do what it suggests by using `pybuild` instead of `python-distutils`. @legoktm are you investigating the second option here (using `pybuild`)? If not, or if you think it'll take some time to investigate, we might want to switch to compat 11 so that we can fix this in the meantime. It looks like 13.7.1 is the latest version of dh, so not sure how version 12 was selected- do you know why? 

(For more info: https://manpages.debian.org/testing/debhelper/debhelper.7.en.html#COMPATIBILITY_LEVELS)

PR #334, opened 2022-05-26
set debhelper compat to 11 for all projects

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-debian-packaging/issues/333

When we bumped the debhelper compat level from 9 to 12, buster and bullseye packages could no longer be built in CI by pointing the builder at a source tarball that was generated by running: `python3 setup.py sdist`. 

We still want to bump the compat version up from 9 to 11 (we only remove `setuptools` in the rules file with this version bump because it's no longer needed). But too much is broken when we bump to 12, so I created an issue to capture the work involved in order to do that bump (https://github.com/freedomofpress/securedrop-debian-packaging/issues/335).

# Test plan

- [x] First confirm that you see the error in both your buster and bullseye build VMs without the fix
- [x] In a bullseye build environment, build a package using `PKG_PATH` but point it to a tarball you created in the project directory using `python3 setup.py sdist` and confirm you can build the package
- [x] In a buster build environment, build a package using `PKG_PATH` but point it to a tarball you created in the project directory using `python3 setup.py sdist` and confirm you can build the package

### DIFF
--- a/securedrop-client/debian/control
+++ b/securedrop-client/debian/control
@@ -2,7 +2,7 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
-Build-Depends: debhelper-compat (= 12), dh-python, python3-all, python3-setuptools, dh-virtualenv
+Build-Depends: debhelper-compat (= 11), dh-python, python3-all, python3-setuptools, dh-virtualenv
 Standards-Version: 3.9.8
 Homepage: https://github.com/freedomofpress/securedrop-client
 X-Python3-Version: >= 3.5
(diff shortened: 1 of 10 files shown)


eaon (approved):
LGTM!

legoktm:
Some of these didn't actually bump the compat version from 9 because the old way of specifying compat was confusing:

old way: in d/control, debhelper (> $compat), plus d/compat with $compat

new way: in d/control, debhelper-compat (= $compat), delete d/compat

Since d/compat wasn't bumped for the old-way packages, those are still on compat level 9. We should take the opportunity to move everything to the new way, it's much simpler.

legoktm:
(if you're interested in the behind the scenes details, https://nthykier.wordpress.com/2018/03/04/prototyping-a-new-packaging-papercut-fix-dry-debhelper-compat-level/ explains the switch to the new way)

creviera:
> Since d/compat wasn't bumped for the old-way packages, those are still on compat level 9. We should take the opportunity to move everything to the new way, it's much simpler.

This was addressed in https://github.com/freedomofpress/securedrop-debian-packaging/pull/338

PR #336, opened 2022-06-03
Rebuild PyYAML wheel with Python 3.9 for bullseye

legoktm:
This is needed by securedrop-proxy.

## Test plan
* [ ] Verify signature is good (CI will spectacularly fail if this is not the case)


creviera (approved):
@legoktm, looks like this is the last dependency in the sha256sum.txt file that needed to have a 3.9 wheel added. Thanks!

PR #338, opened 2022-06-06
Move all packages to debhelper compat 11, use new style of dependency

legoktm:


creviera:
@legoktm - could you take a look at https://github.com/freedomofpress/securedrop-debian-packaging/commit/e88e0bcbce58f070a2046aa7a6a42f3412f9f2ae and see if we missed anything?

creviera:
I see your comment: https://github.com/freedomofpress/securedrop-debian-packaging/pull/334#issuecomment-1147729114

I think it would make sense to update all the projects to the new way in the same PR, would you mind appending to this PR to do that?

legoktm:
Yep

legoktm:
I backed out the securedrop-proxy part because I remembered I did that in the other PR.

creviera (approved):
lgtm!

PR #354, opened 2022-06-30
Stop building securedrop-workstation-grsec package in nightlies

legoktm:
The package is broken and we really don't need nightlies for
kernel things, because the kernel is special.

Fixes https://github.com/freedomofpress/securedrop-workstation/issues/793.


creviera (approved):
lgtm (not a release blocker, but pulling this in)

PR #361, opened 2022-07-13
docs: README correction for changelog script invocation

eloquence:
The script requires a PKG_VERSION as well:

https://github.com/freedomofpress/securedrop-debian-packaging/blob/adf54b816c87fe53f551853d98d6056b1c107486/scripts/update-changelog#L32-L35


cfm (approved):
Looks good, makes sense, matches my own recollection of running `update-changelog`.

PR #349, opened 2022-06-22
Update `securedrop-workstation-config` for bullseye, refrain from building buster nightlies

eaon:
`eog.desktop` was renamed to `org.gnome.eog.desktop` on bullseye.
Updating this is necessary for the client to be able to successfully
open images.

Because the files we want to ship for buster and bullseye now diverge,
and because `securedrop-workstation-config` is unlikely to see a new
release on buster, we want to refrain from building this package on
nightlies, otherwise `securedrop-workstation` tests in dom0 would start
failing on 4.0

Fixes #348

## Testing

On Qubes OS R4.1:

- [ ] (Pre-check) Start a terminal for `sd-viewer`, and run `xdg-mime query default image/jpeg` - it should return `open-in-dvm.desktop`
- [ ] Build `securedrop-workstation-config` manually, transfer it to `sd-large-bullseye-template` and install it with `dpkg -i`
- [ ] Restart `sd-viewer` to make sure it uses the updated package
- [ ] Start a terminal for `sd-viewer`, and run `xdg-mime query default image/jpeg` - it should return `org.gnome.eog.desktop`

PR #371, opened 2022-08-22
Use matrix parameterization to reduce duplication in CI config

legoktm:
* Parameterize nearly all the jobs by package name and distro
* Get rid of all unused common-steps stanzas
* Merge regular and nightly jobs together by conditionally
  determining whether we're supposed to build nightlies or not
* workstation-grsec package left as-is for the most part, it's
  going to be removed from here soon, once <https://github.com/freedomofpress/kernel-builder/pull/24> is merged.

<https://circleci.com/blog/circleci-matrix-jobs/> is a good overview
of the CircleCI features used in this PR.

----
The main motivation for this is that it's now a 4 line PR to add bookworm support!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,16 +16,6 @@ common-steps:
       name: Install Debian packaging dependencies
       command: make install-deps
 
-  - &getlatestreleasedversion
-    run:
-      name: Get latest released version of the project
-      command: |
-        cd ~/packaging/securedrop-*
-        export VERSION_TO_BUILD="$(git describe --tags $(git rev-list --tags --max-count=1))"
-        # Enable access to this env var in subsequent run steps
-        echo $VERSION_TO_BUILD > ~/packaging/sd_version
-        echo 'export VERSION_TO_BUILD=$(cat ~/packaging/sd_version)' >> $BASH_ENV
-
   - &getnightlyversion
     run:
       name: Create nightly version for python packages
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -164,59 +61,25 @@ common-steps:
...
-        echo 'export PKG_NAME=$(cat ~/packaging/sd_package_name)' >> $BASH_ENV
-
   - &setmetapackageversion
     run:
-      name: Get metapackage version via distribution changelog
+      name: Get and set metapackage version via distribution changelog
       command: |
         CURRENT_VERSION=$(grep -oP "\d+\.\d+\.\d+" ${PKG_NAME}/debian/changelog-buster | head -n1)
-        export VERSION_TO_BUILD="$CURRENT_VERSION"
+        if [[ "$SCHEDULE_NAME" != "" || "$CIRCLE_BRANCH" == "pleasebuildnightlies" ]]; then
```
eloquence, line 74: How about using an environment variable like `IS_NIGHTLY` that's `0` on `main` but `1` on `pleasebuildnightlies`, defined here in this configuration itself? That way we're not tied to any specific branch name (but do have to ensure we preserve that flag in `config.yml` on the branch).

.circleci/config.yml
```diff
@@ -723,22 +312,29 @@ workflows:
...
+      - build:
+          matrix:
+            parameters:
+              package:
+                - securedrop-client
+                - securedrop-export
+                - securedrop-log
+                - securedrop-proxy
+              image:
+                - "3.7-buster"
```
eloquence, line 324: Should we just drop this (and the subsequent one now)? We're not using these packages for anything. I realize there's other cleanup to do to move the changelogs over, but I'm not sure there's any point in carrying Buster-related build logic forward at this point.

legoktm: It's mostly to keep me honest and not accidentally introduce any non-parameterized OS behavior. Once this lands, the final step is for me to figure out how to have bookworm jobs OK to fail, but not block the bullseye nightlies. I'm feeling pretty confident that we'll have that done and buster gone by the end of the week.

gonzalo-bulnes (commented):
I'll only add a comment because I'm not familiar at all with this pipelines and @eloquence has reviewed before I did. But I find the resulting matrices excessively readable - to the point that I now understand a little bit more how many and why there are distinct jobs in this config file. :green_heart:

.circleci/config.yml
legoktm: I came up with a better idea at 1am last night that should avoid all of this entirely, stay tuned...

legoktm: OK, there's now a `IS_NIGHTLY` environment variable that will be the string "nightly" during nightly builds, and empty string in regular builds. It's added as a parameter in the nightly workflow, so the job names have "nightly" in them (coming full circle!) and are otherwise identical except for that env variable.

eloquence (approved):
Looks good to me (I've also examined the CI output and the build artifacts generated by your `pleasebuildnightlies` commit on apt-test and did not spot any issues). 

I believe this will reinstate `securedrop-workstation-config` nightlies on buster (disabled via #349) as is also visible on apt-test (compare https://github.com/freedomofpress/securedrop-dev-packages-lfs/commit/66ee129f09b932e294f0aff83a45b64b233ccc0c, generated from this branch, vs. https://github.com/freedomofpress/securedrop-dev-packages-lfs/commit/045120ed7b8faa8c986f783ac21b0975eaa31b0b, generated from `main`). I think that's fine as we're planning to retire all those Buster workflows ASAP anyway, and you're just keeping them around here to ensure the matrix logic is sound and complete.

legoktm:
For reference, https://circleci.com/docs/configuration-reference#excluding-sets-of-parameters-from-a-matrix is how to disable just one part of the matrix, so we could've used that to keep securedrop-workstation-config+buster disabled.

PR #372, opened 2022-08-25
Use plain Debian images in CI

legoktm:
I wrote detailed commit messages for these, so please read those. In summary:
* Document why build-sync-wheels can't be run as root
* Switch to plain Debian images. Fix the reprotest-wheels job so it runs as a non-root user
* Switch lint-and-test and reprotest-debs over to bullseye.


.circleci/config.yml
```diff
@@ -6,15 +6,19 @@ common-steps:
...
+      command: |
+        apt-get update && apt-get install -y make sudo
+        make install-deps
+
+  - &installtestdeps
+    run:
+      name: Install test dependencies
+      command: |
+        sudo apt-get install reprotest faketime -y --no-install-recommends
+        .venv/bin/pip install -r test-requirements.txt
```
gonzalo-bulnes, line 21: Is using `.venv/bin/pip` equivalent to activating the virtual environment?

scripts/build-sync-wheels
```diff
@@ -11,6 +11,12 @@ import argparse
 from pprint import pprint
 
 
+if os.geteuid() == 0:
+    # tar has issues when resetting permissions and ends up
+    # causing very subtle reproducibility issues.
+    print("This script cannot be run as root.")
```
gonzalo-bulnes, line 17: Nitpicking: Would it make sense to output to `STDERR` instead of `STDOUT`? I like the error message otherwise, it's a lot friendlier than letting future us figure out what's going on again!

.circleci/config.yml
```diff
@@ -6,15 +6,19 @@ common-steps:
...
-      command: make install-deps
+      command: |
+        apt-get update && apt-get install -y make sudo
+        make install-deps
+
+  - &installtestdeps
+    run:
+      name: Install test dependencies
+      command: |
+        sudo apt-get install reprotest faketime -y --no-install-recommends
```
eloquence, line 20: I'm curious about the use of `sudo` here and for the `sudo sed` commands -- is that needed here?

gonzalo-bulnes (approved):
I see no blockers. Only left a couple of comments for my own understanding and slight preferences.

.circleci/config.yml
eloquence: It [should be](https://docs.python.org/3/library/venv.html) for `venv`:

> You don’t specifically need to activate an environment; activation just prepends the virtual environment’s binary directory to your path, so that “python” invokes the virtual environment’s Python interpreter and you can run installed scripts without having to use their full path. However, all scripts installed in a virtual environment should be runnable without activating it, and run with the virtual environment’s Python automatically.

.circleci/config.yml
legoktm: No, it's not needed anymore. I think I had left it alone just to reduce the diff, but let's get rid of it.

.circleci/config.yml
legoktm: Yep. Note that there's a subtle gotcha I ran into in the reproducibility tests, which end up shelling out to plain `python`. If you don't explicitly activate the venv, that plain `python` will point to the system Python instead of our venv's. In theory you can work around it by using sys.executable() but I didn't want to go down that route in this PR.

scripts/build-sync-wheels
legoktm: Sure.

gonzalo-bulnes (approved):
Nice, thanks for the explanations @legoktm (and @eloquence). I'm shipping this.

PR #379, opened 2022-09-20
Add alembic 1.1.0 wheel

gonzalo-bulnes:
Rebuilds the securedrop-client wheels for Python 3.10.

https://github.com/freedomofpress/securedrop-client/pull/1555 depends on this.

## Checklist

* [x] `bullseye` builds and jobs are passing
* [x] This PR does not introduce any new `bookworm` test failures

PR #380, opened 2022-09-22
Add pull request templates for common tasks

gonzalo-bulnes:
This is primarily aimed at preventing us from missing any dependency review in the future.

1. Adds an item to remind that diff reviews must be performed when adding packages
2. Adds a template for change log updates, which are relatively frequent
3. Moves the default template, and rewords it slightly to match the other

Prompted by: #379

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/changelog_update.md
+++ b/.github/PULL_REQUEST_TEMPLATE/changelog_update.md
@@ -1,0 +1,9 @@
+---
+name: Change log update
+about: Prepare the release of a Debian package.
+
+---
+
+## Description
+
+_Which package are you releasing? Please link to to **release tracking issue**._
--- a/.github/PULL_REQUEST_TEMPLATE/package_addition.md
+++ b/.github/PULL_REQUEST_TEMPLATE/package_addition.md
@@ -1,0 +1,22 @@
+---
+name: Package addition
+about: Add or upgrade a dependency for one of our Debian packages.
+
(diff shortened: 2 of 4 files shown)


.github/PULL_REQUEST_TEMPLATE/package_addition.md
```diff
@@ -0,0 +1,22 @@
+---
+name: Package addition
```
creviera, line 2: This could also be for a package upgrade. My one request would be that you update wording to include upgrades as well as new packages.

.github/PULL_REQUEST_TEMPLATE/changelog_update.md
```diff
@@ -0,0 +1,9 @@
+---
+name: Change log update
+about: Prepare the release of a Debian package.
+
+---
+
+## Description
+
+_Which package are you releasing? Please link to to **release tracking issue**._
```
creviera, line 9: love it! super helpful

creviera (changes requested):

.github/PULL_REQUEST_TEMPLATE/package_addition.md
gonzalo-bulnes: Good point. I'll add it.

gonzalo-bulnes:
Ready to re-review @creviera !

creviera (approved):
lgtm

PR #384, opened 2022-10-04
Fix a few breaking changes in scripts

legoktm:
* Make sure we call `os.path.expanduser()` on `pkg_dir`
* Try to guess project from `setup.py` instead of expecting it to be cloned in a specific way (that breaks in CI).

Closes https://github.com/freedomofpress/securedrop-client/pull/1572

## Testing
* Rename your `securedrop-client` folder to something else, I did `securedrop-test`.
* from this repo, run `PKG_DIR=~/github/freedomofpress/securedrop-test make requirements`, with a tilde specifically, and make sure it works properly.

### DIFF
--- a/scripts/build-sync-wheels
+++ b/scripts/build-sync-wheels
@@ -7,6 +7,9 @@ import subprocess
 import tempfile
 import shutil
 import argparse
+from pathlib import Path
+
+import utils
 
 
 if os.geteuid() == 0:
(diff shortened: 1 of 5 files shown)


scripts/build-sync-wheels
```diff
@@ -60,13 +62,18 @@ def main():
...
         if not os.path.exists(args.pkg_dir):
             print(f"Project directory missing {args.pkg_dir}.")
             sys.exit(1)
 
     req_path = os.path.join(args.pkg_dir, args.requirements, "requirements.txt")
-    local_wheels = os.path.join(args.project, "wheels")
+    if args.project is not None:
+        project_name = args.project
+    else:
+        project_name = utils.project_name(Path(args.pkg_dir))
```
gonzalo-bulnes, line 75: I'd catch `RuntimeError` (or a more specific one maybe?) here and suggest using the `--project` option. I'm thinking that someone running into that error is very likely to not know about or have forgotten the option.

Something along the lines of: "If you're not creating a Python package, you can use the `--project` option."

gonzalo-bulnes (approved):
Test plan checks out otherwise. :white_check_mark: 

**Edit**: I reviewed ad39bb7 - repeating with 353d415... haha [560338f](https://github.com/freedomofpress/securedrop-debian-packaging/pull/384/commits/560338f7614c351b156ebe1ef5b6fd0ed3e08ed2)!

scripts/build-sync-wheels
legoktm: I adjusted the error message using your suggestion. (and fixed it when setup.py doesn't exist at all). I think this still accomplishes the goal without needing to add an extra try/except to both scripts.

gonzalo-bulnes (approved):
Test plan checks out :white_check_mark:

PR #385, opened 2022-10-04
Update booststrapped tooling

gonzalo-bulnes:
## Description

Updates the bootstrapped tooling as instructed in the README.
**Why now?** I am setting up a more permanent packaging tool chain for myself, and I noticed that running the instructions in the README does update comments in the generated files, that match the changes recently made to the instructions. I think we might as well have all these files updated to avoid future confusions, particularly since the changes are not significant. :slightly_smiling_face: 

## Test plan

- [x] Confirm that the requirements files haven't significantly changed
- [x] Confirm that the comments in the generated files match the [README's instructions](https://github.com/freedomofpress/securedrop-debian-packaging#updating-our-bootstrapped-build-tools)
- [x] Confirm that the hash file hasn't significantly changed
- [x] Confirm that the signature of the hash file is valid

PR #390, opened 2022-10-11
Fix mismatched sha256sums file

gonzalo-bulnes:
## Description

This PR is a follow up on https://github.com/freedomofpress/securedrop-debian-packaging/pull/385
I should have sync'd the hash file when removing the unused wheels, but I didn't notice until I tried updating **click** today.

Unrelated: the copy of my public key happened to have an old expiry date, that just expired. I'm updating the key in order to ensure the signature can be checked as expected.

The update was performed automatically by: `./scripts/sync-sha256sums workstation-bootstrap`. Some hashes moved around, but I assume that's OK because the script is the canonical way to update the hash file.

**Context**: this is a preliminary step to upgrading the **click** dependency.

## Test plan

- [x] Confirm that the hashes that were removed match the wheels that were removed in #385
- [x] Confirm that you can verify the signature of the hash file :lock_with_ink_pen:


gonzalo-bulnes:
I checked that the [instructions to upgrade the bootstrapped tools](https://github.com/freedomofpress/securedrop-debian-packaging/tree/f6225f6e3d2c97c8c3e819e3103b890318268120#updating-our-bootstrapped-build-tools) include the sync'ing step. I don't know how I missed that and didn't notice. :thought_balloon:

gonzalo-bulnes:
:bulb: I know what happened! After removing wheels, the hash files needs to be sync'd again. :slightly_smiling_face:

legoktm (approved):
* [x] Verified that your gpg key has the same fingerprint before and after

PR #396, opened 2022-10-24
Leverage systemd conditionals for logging services

eaon:
Goal of this change is to make it possible to just enable all the services we need on all VMs, but have them start exclusively in the VMs they're meant for.

Since we have predictable hostnames, we can ask systemd to only start services for specific hostnames.

Relies on 2 PRs in other repos:

- https://github.com/freedomofpress/securedrop-workstation/pull/840
- https://github.com/freedomofpress/securedrop-log/pull/34

# Testing

TK


legoktm:
To install systemd units "properly", in d/rules we need:
```
override_dh_installsystemd:
	dh_installsystemd --name=securedrop-disable-remote-logging
```

And the unit should be located at `securedrop-workstation-config/debian/securedrop-workstation-config.securedrop-disable-remote-logging.service`.

Relevant docs at https://manpages.debian.org/bullseye/debhelper/dh_installsystemd.1.en.html#name=

PR #408, opened 2023-01-12
Create issues when a new Tor release is fetched

legoktm:
CI will now create a new issue (or update an existing one) whenever it fetches new Tor packages. This is something that I have been doing manually for a while now whenever I see a new Tor release announcement.

The generated issue contains the checklist as well as the diff of the new debs so you can see the version and checksums of the packages. An example issue (with the wrong patch) is
<https://github.com/freedomofpress/securedrop/issues/6723>.

Some potential future improvements that I deferred for now is to include the version number in the issue title, and the correct Tor Project forum link.

The script wraps around the official `gh` CLI tool, it needs a GITHUB_TOKEN to be set in the environment to work properly. `gh` is only available in bullseye-backports, so I had to adjust the image so it would be installable.

Of course, if this works out well, I'd like to expand this to other things like kernel and dependency updates.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
 
   reprepro-update-tor:
     docker:
-      - image: debian:bullseye
+      - image: debian:bullseye-backports
     steps:
       - checkout
       - *addsshkeys
@@ -174,7 +174,8 @@ jobs:
           name: clone and run reprepro update
           command: |
             apt-get update
-            apt-get install -y reprepro ca-certificates dctrl-tools git git-lfs openssh-client
+            apt-get install -y reprepro ca-certificates dctrl-tools git git-lfs openssh-client \
+                gh python3
 
             # Clone the dev repo and configure it
             git clone git@github.com:freedomofpress/securedrop-dev-packages-lfs.git
@@ -193,7 +194,8 @@ jobs:
             mv repo/pool/main/t/tor/*.deb core/focal/
             git add core/focal/*.deb
             # If there are changes, diff-index will fail, so we commit and push
-            git diff-index --quiet HEAD || git commit -m "Automatically updating Tor packages" && git push origin main
+            git diff-index --quiet HEAD || git commit -m "Automatically updating Tor packages" \
+                && git push origin main && ../scripts/new-tor-issue
 
 
   build: &build
--- a/scripts/new-tor-issue
+++ b/scripts/new-tor-issue
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+"""
+Creates a new issue to track Tor updates or appends a comment
+to any existing issues
+"""
+
+import json
+import random
+import subprocess
+import tempfile
+
+REPOSITORY = "freedomofpress/securedrop"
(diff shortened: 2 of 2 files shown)


legoktm:
Token has been created, so this is ready for review. I will also kick off a test job tomorrow to make sure the token is read properly.

legoktm:
I pushed https://github.com/freedomofpress/securedrop-builder/commit/50ea9e47a81a1c4d9b00b8c8e07aeab5644fc960 to force run the issue creation script, but it failed with:
```
HTTP 401: Personal access tokens with fine grained access do not support the GraphQL API (https://api.github.com/graphql)
Try authenticating with:  gh auth login
```

Per https://github.com/community/community/discussions/36441 this is a known limitation, I've asked infra if we can switch to a classic token for now.

legoktm:
The new classic token works! It created https://github.com/freedomofpress/securedrop/issues/6730.

scripts/new-tor-issue
```diff
@@ -0,0 +1,70 @@
...
+</details>
+
+* [x] CI adds new packages to apt-test
+* [ ] Install tor, tor-geoipdb packages from apt-test on a prod
+      install and let them sit overnight
+* [ ] Verify that tor is still running after reboot, services
+      available, no errors or unexpected messages in logs
+* [ ] Submit a PR to `securedrop-debian-packages-lfs` to deploy
+      the same packages
+"""
```
cfm, line 37: How would you feel about having the new ticket point to where it came from, so that the automation is self-documenting?

```suggestion

PS.  This pull request was created by `scripts/new-tor-issue` via CI job `reprepro-update-tor`.
"""
```

cfm (changes requested):
Looks great to me, @legoktm.  Left one suggestion inline, to keep us sane if we do experiment with more automations like this one.  :-)

scripts/new-tor-issue
legoktm: Good idea! I'll push it in a minute, with `s/pull request/issue/`.

cfm (approved):
Thanks, @legoktm!

.circleci/config.yml
```diff
@@ -193,7 +194,8 @@ jobs:
             mv repo/pool/main/t/tor/*.deb core/focal/
             git add core/focal/*.deb
             # If there are changes, diff-index will fail, so we commit and push
-            git diff-index --quiet HEAD || git commit -m "Automatically updating Tor packages" && git push origin main
+            git diff-index --quiet HEAD || git commit -m "Automatically updating Tor packages" \
```
legoktm, line 197: This did not work like I expected (and already hasn't been working like I expected).

```
[user@dev ~]$ false || echo "1" && echo "2" && echo "3"
1
2
3
[user@dev ~]$ true || echo "1" && echo "2" && echo "3"
2
3
```

Or with parenthesis, I assumed/wanted `diff-index || (commit && push && new-tor-issue)` but instead we got `(diff-index || commit) && push && new-tor-issue`.

Which means it'll always run the script, regardless of whether there's a new package :<

PR #409, opened 2023-01-12
Have CI build nightlies for securedrop-updater

legoktm:
In theory it's general but currently expecting the pending securedrop-updater Makefile and folder layout and assume that securedrop-workstation's rpm will be modified to follow this layout too.

Fixes https://github.com/freedomofpress/securedrop-updater/issues/10

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,6 +273,47 @@ jobs:
     environment:
       CODENAME: bookworm
 
+  build-rpm:
+    parameters:
+      package:
+        type: string
+    docker:
+      - image: fedora:32
+    environment:
+      PKG_NAME: << parameters.package >>
+    steps:
+      - checkout
+      - *addsshkeys
+      - run:
+          name: Clone and install dependencies
+          command: |
+            dnf install git git-lfs make -y
+            git clone https://github.com/freedomofpress/${PKG_NAME}.git
+            cd ${PKG_NAME}
+            make install-deps
+      - run:
+          name: Bump version and build rpm
+          command: |
+            cd ${PKG_NAME}
+            # Version format is "${VERSION}-0.YYYYMMDDHHMMSS.fXX", which sorts lower than "${VERSION}-1"
+            rpmdev-bumpspec --new="$(cat VERSION)-0.$(date +%Y%m%d%H%M%S)%{?dist}" rpm-build/SPECS/${PKG_NAME}.spec
+            make build-rpm
+      - run:
+          name: Commit and push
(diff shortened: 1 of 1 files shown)


legoktm:
Also the infra scripts need to be updated to handle a `-nightlies` folder.

legoktm:
The rpm test signing key (`4ED79CC3362D7D12837046024A3BE4A92211B03C`) is currently kept private and individual developers have copies of it. rpm packages are individually signed rather than the repository metadata (apt/dpkg), so we need to figure out how CI/automation is expected to sign the packages. Just listing out the options:

1) CircleCI has access to the private key and signs the packages. We currently use CircleCI to build the packages, so we already trust CircleCI at some level to not tamper with them.
1b) CircleCI does not have a recent good security track record, so maybe we use GitHub Actions (or some other host) for buliding and give that access to the private key.
2) Packages in the repository are unsigned, but the automation on yum-test.sd.o signs the rpms before generating repodata. This is similar to the apt-test workflow, where the private key lives on the server and it signs the repo metadata before publishing.
3) ??

An attacker with access to the private key would also need push access to the securedrop-workstation-dev-rpm-packages-lfs repo (or access to the yum-test server) to publish malicious packages, which would be installed into dom0 of developer laptops. If we had nightlies with autosigning, presumably it would be easier to just get malicious code into a github repo, which then gets built and pushed as a nightly.

eloquence:
> Packages in the repository are unsigned, but the automation on yum-test.sd.o signs the rpms before generating repodata. This is similar to the apt-test workflow, where the private key lives on the server and it signs the repo metadata before publishing.

This seems like a reasonable option from my end, as it largely replicates the security model we're already using for another package format. If you, @l3th3 and others agree, I would suggest opening an infra issue to that effect -- perhaps this could be an opportunity for direct collaboration with the infra team on that codebase.

legoktm:
OK, we're all set on the infra side so @cfm this is ready for your review :)

We also need to land https://github.com/freedomofpress/securedrop-yum-test/pull/44 - I think that will require admin access (@eloquence?) to remove the project from CircleCI first.

eloquence:
Done. The steps were:
1. Instruct CircleCI to stop building the repo via CircleCI project settings (CircleCI inherits org-level admin permissions)
2. Amend the commit to ensure that checks are re-run.

cfm (approved):
Very nice!  The `nightly` workflow won't run on this branch, and I think testing locally with `circleci` is both overkill and secrets-deficient, but this looks great to me.

PR #415, opened 2023-02-15
Make nightly versions compliant with PEP 440

### LINKED ISSUE #414: bookworm nightlies failling because version number is now invalid
```
  /src/build/debbuild/packaging/securedrop-proxy/debian/securedrop-proxy/opt/venvs/securedrop-proxy/lib/python3.11/site-packages/setuptools/dist.py:548: UserWarning: The version specified ('0.4.0-dev-20230214-235645') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
```

From https://peps.python.org/pep-0440/#public-version-identifiers

> The canonical public version identifiers MUST comply with the following scheme:
> `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`

Where `N` is one of `[a-zA-Z]`, `[0-9]` or a literal `.`

I think we can adjust our version format to `0.4.0.dev20230214235645`, which is the same value just without dashes.

This version is also used by dpkg, but we should get the same semantics as before there:
```
# dpkg --compare-versions '0.4.0-dev-20230214-235645' '>=' '0.4.0' && echo "correct"
correct
# dpkg --compare-versions '0.4.0.dev20230214235645' '>=' '0.4.0' && echo "correct"
correct
```

legoktm:
The setuptools in bookworm now enforces PEP 440 compliance for version numbers rather than just warning about it.

We can fall into compliance by just removing the extra dashes.

Specifically:
 Previous format: 0.4.0-dev-20230214-235645
 New format: 0.4.0.dev20230214235645

Fixes #414.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,8 @@ common-steps:
       name: Create nightly version for python packages
       command: |
         cd ~/packaging/securedrop-*
-        # Nightly versioning format is: LATEST_TAG-dev-YYMMDD-HHMMSS
-        export VERSION_TO_BUILD="$(git describe --tags $(git rev-list --tags --max-count=1))-dev-$(date +%Y%m%d)-$(date +%H%M%S)"
+        # Nightly versioning format is: LATEST_TAG.devYYMMDDHHMMSS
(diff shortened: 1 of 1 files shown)


rocodes (approved):
Thank you @legoktm :)

PR #416, opened 2023-02-16
Stop patching dh_virtualenv for Python 3.11+, now fixed

legoktm:
The fixed dh_virtualenv version is now in Debian Bookworm, so we no longer need to patch it ourselves:
* https://tracker.debian.org/news/1418871/accepted-dh-virtualenv-122-13-source-into-unstable/
* https://tracker.debian.org/news/1420883/dh-virtualenv-122-13-migrated-to-testing/

## Test plan
* [x] CI passes

### DIFF
--- a/scripts/install-deps
+++ b/scripts/install-deps
@@ -24,10 +24,6 @@ sudo apt-get install  \
     python3-setuptools \
     desktop-file-utils -y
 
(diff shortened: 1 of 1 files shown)


cfm (approved):
Thanks, @legoktm!

PR #417, opened 2023-02-21
Remove dependencies no longer required by securedrop-proxy

legoktm:
See 24cfed80fb26 and afc33dcfe9f08 in that repository.


eaon (approved):
Yusss

PR #419, opened 2023-03-09
Updates for securedrop-client 0.9.0-rc1 build

zenmonkeykstop:
- updates scripts to reference `changelog-bullseye` instead of `changelog-buster`
- adds changelog entry for securedrop-client 0.9.0-rc1


legoktm:
I would leave it as `changelog-buster` for now, the build script is smart enough to rewrite it to bullseye as necessary. (My plan was to rename them to plain `changelog` when we did https://github.com/freedomofpress/securedrop-builder/issues/308 but haven't gotten around to that just yet...)

zenmonkeykstop:
OK, just redoing it in that case.

PR #420, opened 2023-03-09
Update Kunal's public key

legoktm:
The expiry has been extended to 2024-06-14. The fingerprint has not changed, otherwise it can be verified against the key posted on my website[1], keys.openpgp.org, etc.

[1] https://legoktm.com/w/index.php?title=PGP


zenmonkeykstop (approved):
FP looks good,  :+1:

PR #423, opened 2023-03-10
Add script to check newly added packages against buildinfo files

legoktm:
buildinfo files contain package checksums in a machine-readable format, so script checking newly added packages against those.

This will be added to CI for securedrop-apt-test and securedrop-apt-prod.

The main iffy part of this is how it compares against "origin/main", but I think for PRs it'll mostly do the right thing. We only check new packages because old ones don't have buildinfo published. Maybe once we no longer have any legacy cases left, we just check everything in the repository.

Likely there are more checks that could be added, but this is a start.

Refs <https://github.com/freedomofpress/securedrop/issues/6356>.

### DIFF
--- a/scripts/check-buildinfo
+++ b/scripts/check-buildinfo
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+"""
+Check newly added packages against their buildinfo files
+
+Example:
+    ./check-buildinfo package.deb package.buildinfo
+
+"""
+import argparse
+import hashlib
+import subprocess
+import sys
+from pathlib import Path
+from typing import List
+
+from debian.deb822 import BuildInfo
+
+
+def lookup_buildinfos(buildinfos: Path) -> dict:
+    """Extract checksums out of every buildinfo file we can find"""
+    data = {}
+    for path in buildinfos.glob("**/*.buildinfo"):
+        info = BuildInfo(path.read_text())
+        for details in info['Checksums-Sha256']:
+            if details['name'].endswith('.deb'):
+                data[details['name']] = details['sha256']
+    return data
+
+
+def check_package(package: Path, buildinfos: dict) -> bool:
+    """Verify the package's checksum matches buildinfo"""
+    try:
+        expected = buildinfos[str(package.name)]
+    except KeyError:
+        print(f"ERROR: Unable to find buildinfo containing {package.name}")
+        return False
+    actual = hashlib.sha256(package.read_bytes()).hexdigest()
+    if actual == expected:
+        print(f"OK: got expected checksum {actual} for {package.name}")
+        return True
+    else:
+        print(f"ERROR: package is {actual}, buildinfo has {expected} for {package.name}")
+        return False
+
+
+def added_files(against="origin/main") -> List[Path]:
+    """Get list of added files compared to main"""
+    added = []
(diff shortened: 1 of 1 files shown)


legoktm:
https://github.com/freedomofpress/securedrop-apt-test/pull/188 shows the script in action in CI against the new Linux packages. From the log:
```
OK: got expected checksum 58c91237ec828dd058730329c4530c597f22edf96f8a4013027bcc622f37c968 for linux-headers-5.15.98-1-grsec-securedrop_5.15.98-1-grsec-securedrop-1_amd64.deb
OK: got expected checksum cb32a3331314c3bc499428c54fc5bc633aac101baac85000ecf1d64f7ebdf51f for linux-headers-5.15.98-1-grsec-workstation_5.15.98-1-grsec-workstation-1_amd64.deb
OK: got expected checksum 55183b23009ae59ecf289e2d4a8016b97b861ec6877afab8a0089603d830e831 for linux-image-5.15.98-1-grsec-securedrop_5.15.98-1-grsec-securedrop-1_amd64.deb
OK: got expected checksum b9c0e2b85269e03989f63623cb87d90e5bd73b4e6d3b01caf6066e57e3973507 for linux-image-5.15.98-1-grsec-workstation_5.15.98-1-grsec-workstation-1_amd64.deb
OK: got expected checksum f2d50860cef1a4ee898bac57b4d5f3c041192d922c420d6593461c3768870db8 for securedrop-grsec_5.15.98-1-grsec-securedrop-1_amd64.deb
OK: got expected checksum e3c87f843af381c32fdbd98614706b8bb2df8d8341e6bd7b1646e359e7bbadba for securedrop-workstation-grsec_5.15.98-1-grsec-workstation-1_amd64.deb
```

scripts/check-buildinfo
```diff
@@ -0,0 +1,89 @@
...
+
+
+def lookup_buildinfos(buildinfos: Path) -> dict:
+    """Extract checksums out of every buildinfo file we can find"""
+    data = {}
+    for path in buildinfos.glob("**/*.buildinfo"):
+        info = BuildInfo(path.read_text())
+        for details in info['Checksums-Sha256']:
+            if details['name'].endswith('.deb'):
+                data[details['name']] = details['sha256']
```
cfm, line 26: Would it be a useful sanity-check to make sure we haven't already seen a checksum for `details['name']`?

```suggestion
            if details['name'].endswith('.deb'):
                if details['name'] in data:
                    raise ValueError(f"ERROR: duplicate checksum for {details['name']}")
                data[details['name']] = details['sha256']
```

cfm (changes requested):
Very nice.  I've left one paranoiac question for you, @legoktm, but otherwise I'm happy to approve and merge.

scripts/check-buildinfo
legoktm: I think that's a good idea just as general hardening.

In theory people will commit the buildinfo files using the generated filename, so there shouldn't be room for duplicates, but it's also possible we'll move packages around (e.g. securedrop-grsec was formerly built by securedrop server, now it's built by the kernel stuff) and accidentally cause duplicates. An attacker could theoretically also add a buildinfo with an easy-to-overlook extra entry that clobbers the correct checksum with their own...so overall, good idea!

PR #424, opened 2023-03-13
Add build changes for securedrop-client 0.9.0-rc2

zenmonkeykstop:
- debian changelog entry for 0.9.0-rc2 added
- `files/securedrop-client.desktop` added back to debian .install file (see securedrop-client#1638)


zenmonkeykstop:
This is gonna cause a fun chicken and egg sitch if the .desktop file is not added back on main as well...

rocodes (approved):

PR #427, opened 2023-03-14
update build script to remove PKG_VERSION, use only PKG_PATH or PKG_GITREF modes.

### LINKED ISSUE #426: The Debian build script outputs incorrect log messages when using PKG_PATH option
The `scripts/build-debianpackage` script can be used in two ways:

- if the `PKG_PATH` env var is set, it will attempt to build a debian package from either a source tree or a source tarball at the location in PKG_PATH, using the latest changelog entry to set the package version
- if `PKG_PATH` is not set, but `PKG_VERSION` is set, it will attempt to checkout and verify a production-signed tag `$PKG_VERSION` and build the debian package from the tagged source.

In recent builds, it came to light that even if it was invoked with PKG_PATH, the script will still set PKG_VERSION  using the latest tag and either fail if that tag is not production-signed, or just use PKG_VERSION erroneously in messages instead of the actual version as pulled from the changelog.

This causes confusion, and delay.

zenmonkeykstop:
Fixes #426 

Removes option to build using PKG_VERSION. Build options are now PKG_PATH and PKG_GITREF:

- when invoked without either PKG_PATH or GITREF set, will error out (previously built last known release)
- when invoked with PKG_PATH, will use the source tree or tarball at `$PKG_PATH` to build package (no checkout, no tag verification)
- when invoked with PKG_GITREF, will check out corresponding gitref attempt release key verification verification if PKG_GITREF looks like a release tag (catching sneaky branches!) and, if all is well, build package 

In all cases the package version will be derived from the latest version in the source tree's `debian/changelog` If a release tag PKG_GITREF is specified and doesn't match the changelog version, the build will not proceed.

## Testing:

- [x] test building an old prod tag `X.Y.Z` with `PKG_GITREF=X.Y.Z scripts/build-debianpackage`
- [x] test building a non-prod tag or branch with `PKG_GITREF=<ref here> scripts/build-debianpackage`
- [x] test building an existing source tree with `PKG_PATH=../securedrop-foo scripts/build-debianpackage`

### DIFF
--- a/scripts/build-debianpackage
+++ b/scripts/build-debianpackage
@@ -1,11 +1,14 @@
 #!/bin/bash
 # Wrapper script to build Debian packages for the SecureDrop Workstation.
-# Requires a source tarball to build the package, defined via `PKG_PATH`.
-# The script will check for a suitable tarball on an adjacent directory.
+# Requires a source tarball to build the package, defined via `PKG_PATH`,
+# or a git reference defined via `PKG_GITREF`. If `PKG_GITREF` matches the
+# release tag format, it will be treated as a tag and expected to be signed
+# with the production release key.
+#
 # Explicit configuration is available via the following env vars:
 #
 #   * PKG_PATH
-#   * PKG_VERSION
+#   * PKG_GITREF
 #   * PKG_NAME
 #
 set -e
@@ -17,6 +20,13 @@ set -o pipefail
 if [[ -z "${PKG_NAME:-}" ]]; then
     echo "Set PKG_NAME of the build";
     exit 1
+fi
+
+if [[ -z "${PKG_PATH:-}" ]]; then
+    if  [[ -z "${PKG_GITREF:-}" ]]; then
+        echo "Set PKG_PATH or PKG_GITREF";
+        exit 1
+    fi
 fi
 
 # Store root of repo, since we'll change dirs several times.
@@ -40,28 +50,6 @@ mkdir -p "$TOP_BUILDDIR"
 rm -rf "${TOP_BUILDDIR:?}/${PKG_NAME}"
 mkdir -p "${TOP_BUILDDIR}/${PKG_NAME}"
 
-
-# Look up most recent release from GitHub repo
-function find_latest_version() {
-    repo_url="https://github.com/freedomofpress/${PKG_NAME}/tags"
-    curl -s "$repo_url" \
-        | perl -nE '$_ =~ m#/releases/tag/(v?[\d\.]+)\"# and say $1' \
-        | head -n 1
-}
-
-# If not building from an existing source tree AND not specifying
-# a prod release version, grab the latest prod tag.
-if [[ -z "${PKG_VERSION:-}" ]] && [[ -z "${PKG_PATH:-}" ]]; then
-    echo "PKG_VERSION not set, inferring from recent releases..."
-    PKG_VERSION="$(find_latest_version)"
-    if [[ -z "$PKG_VERSION" ]]; then
-        echo "Failed to infer version"
-        exit 1
-    else
-        echo "Using PKG_VERSION: $PKG_VERSION"
-    fi
-fi
-
 # Ensures that a given git tag is signed with the prod release key
 # If "rc" is in the tag name, this will fail.
 function verify_git_tag() {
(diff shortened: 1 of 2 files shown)


legoktm:
Nothing wrong with your code from a quick skim, but I'm not really sure we need to keep PKG_VERSION around? It currently does only two things 1) verify prod signed tag 2) print out what version we're using.

No. 2 is duplicative of dpkg-buildpackage's own message ("dpkg-buildpackage: info: source package securedrop-client
" & "dpkg-buildpackage: info: source version 0.9.0-rc2+bullseye").

So that just leaves No 1., verifying the prod signed tag. And I feel like there's a better way to do that... for example, we could run `git tag --points-at HEAD` and see whether it prints a `\d+?\.\d+?\.\d+?` matching string and if so, verify against the prod key.

zenmonkeykstop:
I'd go so far as to say (2) is actually misleading (hence the PR) - the package version is determined by the changelog, not by the tagged version that gets checked out. IMO we still have 3 basic cases to cover:
- like, build whatever source tree we get pointed at (wanna build a branch? nightlies? wanna build an rc? this is for you)
- check out an arbitrary ref and build it (wanna build as above without the hassle of prepping a source tree? et voila)
- check out a ref we expect to have a prod-signed tag and if so, build it (ok this is a serious build, let's be duly diligent)

But maybe what this actually points to is that we should be decomposing this script, doing the source tree prep and optional verification elsewhere if required, always just providing the prepped source tree, and focusing on the debian build bits.

legoktm:
> I'd go so far as to say (2) is actually misleading (hence the PR) - the package version is determined by the changelog, not by the tagged version that gets checked out. 

Agreed.

> IMO we still have 3 basic cases to cover:
> 
> * like, build whatever source tree we get pointed at (wanna build a branch? nightlies? wanna build an rc? this is for you)
> 
>  * check out an arbitrary ref and build it (wanna build as above without the hassle of prepping a source tree? et voila)
> 
>   * check out a ref we expect to have a prod-signed tag and if so, build it (ok this is a serious build, let's be duly diligent)

The third bullet is really just a special case of the second, no? In the MediaWiki release script that's basically how I treated it - accept any git ref, but if that ref is a tag, perform extra checks: https://gitlab.wikimedia.org/repos/releng/release/-/blob/f9857bd33d694aa6c492372f4111f599a3df6c87/make-release/makerelease2.py#L82

> But maybe what this actually points to is that we should be decomposing this script, doing the source tree prep and optional verification elsewhere if required, always just providing the prepped source tree, and focusing on the debian build bits.

"Yes, but..." I think that would be the ideal state to end up in, but I think we right now don't have a lot of confidence/trust in our builds yet because it is convoluted that I think having just one thing to run for now is better and once the whole process is simpler we can start breaking it down a bit.

zenmonkeykstop:
OK this is probably in better shape if @legoktm or anyone else has :eyes: time. PKG_VERSION dropped altogether, some hip-height guardrails added around production builds.

scripts/build-debianpackage
```diff
@@ -100,9 +90,11 @@ if [[ "${PKG_NAME}" =~ ^(securedrop-client|securedrop-proxy|securedrop-export|se
 
     if [[ -z "${PKG_PATH:-}" ]]; then
         # Build from source
-        echo "PKG_PATH not set, building from source (version $PKG_VERSION)..."
-        setup_source_tree
-        PKG_PATH="/tmp/${PKG_NAME}/"
+        if [[ -n "${PKG_GITREF:-}" ]]; then
+            echo "PKG_PATH not set, using git reference $PKG_GITREF)..."
```
legoktm, line 94: Unnecessary ) ?

legoktm (changes requested):
Minor thing, otherwise looks good. I do think we have officially reached the limits of what we can safely + easily do in bash scripting and any more complexity likely needs a Python port first.

zenmonkeykstop:
> Minor thing, otherwise looks good. I do think we have officially reached the limits of what we can safely + easily do in bash scripting and any more complexity likely needs a Python port first.

I'd agree with this assessment, was hard to resist the temptation.

scripts/build-debianpackage
zenmonkeykstop: Nuked!

legoktm (approved):
LGTM, test plan checks out. Docs PR is in progress, don't think we need to block on that.

PR #433, opened 2023-03-24
WIP: Script to rebuild a package from buildinfo

legoktm:


legoktm:
rebuilding securedrop-client 0.9.0 with this basically worked: https://gist.github.com/legoktm/3352b54b373bc9f186f555a087a6d3a4 - all the diffs are known issues.

Remaining work:
* We need to capture the git commit sha1 of the repository being packaged as well as securedrop-builder.
* Presumably the correct order of operations is something like:
   * parse buildinfo, check out correct version of securedrop-builder (in place? somewhere else?)
   * Run `make install-deps` to install apt packages and bootstrap
   * Set up snapshot apt config, force install correct versions of packages (hope that python, etc. didn't materially change and the bootstrap doesn't need to be recreated??)
   * Build the package
   * Print diff of the buildinfo (at least build date should be different)
   * Print diffoscope of the debs (ideally no differences)

I don't think we're yet at the point of automating this, but we should try it during release time to verify the buildinfo/package.

In https://github.com/freedomofpress/securedrop/issues/6356#issuecomment-1470566604 I explained how we can add in more environment variables. I think we should prefix them with "SD_", so we probably want to rename `PKG_GITREF`, etc. to use that. And then probably some new `SD_BUILDER_GIT_VERSION`

legoktm:
This is outdated and not really useful. We have pretty reliable reproducible builds that don't need a buildinfo and we're happy with that.

PR #434, opened 2023-03-28
Add SecureDrop-specific metadata to buildinfo files

legoktm:
dpkg will only export specific known environment variables into the
buildinfo file[1], but we want to track at least two more things:
    
1) the Git commit of the package being built
2) the Git commit of the securedrop-builder repository being used, as 
    the bootstrap and wheels used affect the package output.
    
We can track those by capturing the values before the build process and
then manually adding them to the end of the buildinfo file. This also
means that builds must be done from a Git checkout, and cannot be from a
tarball, so that path now errors out.
    
While we're at it, look up the package filename by "parsing"
debian/files instead of trying to find it via find.
    
[1] https://git.dpkg.org/cgit/dpkg/dpkg.git/tree/scripts/Dpkg/BuildInfo.pm

I am also renaming `PKG_GITREF` to `SD_PKG_GITREF`, as this will be exported in the `.buildinfo` file so let's prefix it with "SD" to make it abundantly clear this is our environment variable and reduce the risk of collision.

There don't appear to be any other users of this besides humans, so I have not added in any backwards-compatibility support for the old name.

## Test plan
* [ ] Check out this PR. Run `SD_PKG_GITREF=main make securedrop-proxy`, verify the buildinfo file contains SD_BUILDER_GITREF and SD_PKG_GITREF with commits that correspond to this builder PR you just checked out and the current main branch of sd-proxy.
* [ ] Add a dummy commit to your sd-builder checkout. Re-run `SD_PKG_GITREF=main make securedrop-proxy`, verify SD_BUILDER_GITREF changed to correspond to the new commit you just added.
* [ ] Run `git -C /tmp/securedrop-proxy checkout HEAD~1`, note which commit you're now on. Run `PKG_PATH=/tmp/securedrop-proxy make securedrop-proxy`, verify SD_PKG_GITREF changed to the new commit you switched to.
* [ ] Run `PKG_VERSION=1 make securedrop-keyring`, verify the buildinfo file contains SD_BUILDER_GITREF but not SD_PKG_GITREF.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,13 @@
     ```shell
     # From a release tag x.y.z signed by the SecureDrop Release Signing key
     PKG_VERSION=x.y.z ./scripts/update-changelog securedrop-client
-    PKG_GITREF=x.y.z make securedrop-client
+    SD_PKG_GITREF=x.y.z make securedrop-client
     ```
     
     ```shell
     # From a non-release tag or branch
     PKG_VERSION=<version> ./scripts/update-changelog securedrop-client
-    PKG_GITREF=<ref> make securedrop-client
+    SD_PKG_GITREF=<ref> make securedrop-client
     ```
     
     ```shell
@@ -166,4 +166,4 @@ Finally, submit a PR containing the new wheels and updated files.
 If you wish to test the new wheels in a local build before submitting a PR,
 or as part of PR review, you can do so by:
 
-Then run e.g. `PKG_GITREF=0.4.1 make securedrop-client` to verify that the new wheels are working.
+Then run e.g. `SD_PKG_GITREF=0.4.1 make securedrop-client` to verify that the new wheels are working.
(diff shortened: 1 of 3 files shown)


legoktm:
One open question that I kind of discussed in https://github.com/freedomofpress/securedrop-builder/pull/433#issuecomment-1483382710 is at what layer should `SD_BUILDER_GITREF` be handled. It seems absurdly complex to try and have the build-debianpackage script re-clone the builder repo if it's the wrong version, my current thinking is we do it one layer higher and have the rebuild script take care of it.

But then should the build-debianpackage script error out if SD_BUILDER_GITREF is set and we're using the wrong builder? I think that would be a reasonable fail-safe.

legoktm:
This is basically ready for review, but lets land https://github.com/freedomofpress/securedrop-builder/pull/427 first and then I'll rebase this.

cfm:
@legoktm and I discussed this in the DX working group just now.  @zenmonkeykstop, I can pick this up, unless you're attached to it.

legoktm:
I need to port this over to the securedrop-client repository.

legoktm:
Still probably useful, but needs to be ported over to the client repo.

PR #452, opened 2023-06-01
Remove Debian autogenerated conffiles from securedrop-keyring

### LINKED ISSUE #450: Clean up debian-generated files in `/etc`, particularly in `/etc/apt/trusted.gpg.d`
Noticed during "QA" for https://github.com/freedomofpress/securedrop-builder/issues/443: When we install a debian package that overwrites files in `/etc/`, such as `/etc/apt/trusted.gpg.d/securedrop-keyring.gpg`, Debian keeps the old file and adds new files (`/etc/apt/trusted.gpg.d/securedrop-keyring.gpg~`, the old key, and `/etc/apt/trusted.gpg.d/securedrop-keyring.gpg.dpkg-dist`, which has the new key).

This is normal, but we should decide on how to handle cleanup for these files.

rocodes:
Closes #450 

- Don't include conffiles for securedrop-keyring 
- Add deb test that tests this behaviour only for securedrop-keyring, and add test to `make tests` target

### Test plan
- In a debian-based vm, install seuredrop-keyring-0.1.7 from apt-test
- Build securedrop-keyring-0.2.0 deb from the tip of this branch, and install it using dpkg
- [x] deb installs successfully and no additional files are created in `/etc/apt/trusted.gpg.d/`
- [x] bonus: importing the new keyring to an as-yet-unsed gpg homedir (eg `gpg --homedir /tmp --import /etc/apt/trusted.gpg.d/securedrop-keyring.gpg`) shows only the 2024 expiry pubkey

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ build-wheels: ## Builds the wheels and adds them to the localwheels directory
 
 .PHONY: test
 test: ## Run simple test suite (skips reproducibility checks)
-	pytest -v tests/test_update_requirements.py
+	pytest -v tests/test_update_requirements.py tests/test_deb_package.py
 
 .PHONY: reprotest
 reprotest: ## Runs only reproducibility tests, for .deb and .whl files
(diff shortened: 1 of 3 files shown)


legoktm (approved):
Awesome, test plan checks out!

PR #454, opened 2023-06-21
Add requests-2.31.0 wheel to securedrop-proxy

rocodes:
Towards https://github.com/freedomofpress/securedrop-proxy/pull/119

Add local wheel for requests-2.31.0 and update shasums

- [x] Detached signature is verified
- [x] Diff review is present (https://github.com/freedomofpress/securedrop-builder/wiki/requests-2.26.0-to-2.31.0)
- [x] Reproducible hashes for build artifacts [note: I have verified by doing build twice in dvm so this is at reviewer's discretion]
- [x] `make securedrop-proxy` completes and updated dependency is seen in console output


legoktm (approved):
Test plan checks out

PR #461, opened 2023-08-02
Add script to automate creation of monthly kernel issues

legoktm:
Per <https://xkcd.com/1205/> this will shave off about a minute on a monthly task, and it took me less than an hour so it was probably worth the time.


legoktm:
This is no longer needed and was integrated into the kernel test farm automation.

PR #467, opened 2023-09-20
Add support for building packages that use Poetry

eloquence:
This draft PR adds optional support for building packages that use Poetry. It includes some other minor changes:
- comment/documentation updates to avoid "PyPI mirror" terminology in favor of simpler "local wheels" terminology (and some corrections to update us with recent changes in this repo)
- avoid use of deprecated `imp` module in tests
- consolidate utility functions for easier maintenance; add tests including for previously uncovered code

It pulls in @legoktm's work in the `wip/poetry` branch. 

Some remaining TODOs on my plate:
- Finish up the `only-poetry` branch in `securedrop-proxy` so it can be used to accompany and test this PR
- Perform end-to-end manual testing of all supported functionality
- Write a test plan for this PR

I'm opening this in draft form to invite overall comments on organization and approach, or stuff I may be missing. :)


legoktm (changes requested):
This is all great, I checked out the `only-poetry` branch, then ran `poetry update` (pinning an older idna for now) and was able to build new wheels and then update the build-requirements.txt file!!

I left some code suggestions inline. The only other comment I have is whether it makes sense to keep the `requirements/` folder since it'll now contain just one item, the `build-requirements.txt` file. I think we could move it into the root now?

scripts/build-sync-wheels
```diff
@@ -19,9 +19,8 @@ if os.geteuid() == 0:
     # sys.exit(1)
 if "VIRTUAL_ENV" not in os.environ:
     print(
-        "This script should be run in a virtualenv: "
-        "`source .venv/bin/activate`",
-        file=sys.stderr
+        "This script should be run in a virtualenv: " "`source .venv/bin/activate`",
```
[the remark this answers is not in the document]
eloquence: Done in 6754fac38bfa27e77ab22c5b543f640010842083

scripts/utils.py
```diff
@@ -2,19 +2,176 @@
...
+
+
+def get_poetry_hashes(
+    path_to_poetry_lock: Path, path_to_pyproject_toml: Path
+) -> dict[str, list[str]]:
+    """
+    Get a dictionary for all main (non-development) dependencies and their
+    valid hashes as defined in a set of requirements as defined in
+    pyproject.toml/poetry.lock. This includes transitive dependencies of
+    main depenencies.
```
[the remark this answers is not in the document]
eloquence: Fixed in 6754fac38bfa27e77ab22c5b543f640010842083

tests/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    expected_output = ["colorama", "cowsay", "beautifulsoup4", "soupsieve"]
+    output = get_relevant_poetry_dependencies(PYPROJECT_TOML_PATH, POETRY_LOCK_PATH)
+    # Ensure we correctly ignore development-only dependencies
+    assert "pytest" not in output
+    assert sorted(output) == sorted(expected_output)
+
+
+def test_get_poetry_hashes():
+    output = get_poetry_hashes(POETRY_LOCK_PATH, PYPROJECT_TOML_PATH)
+    assert list(output.keys()) == EXPECTED_KEYS
```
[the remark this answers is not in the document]
eloquence: Fixed in 6754fac38bfa27e77ab22c5b543f640010842083

tests/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    assert all(len(output[dep]) > 0 for dep in output.keys())
```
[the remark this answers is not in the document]
eloquence: Fixed in 6754fac38bfa27e77ab22c5b543f640010842083

eloquence:
> The only other comment I have is whether it makes sense to keep the requirements/ folder since it'll now contain just one item, the build-requirements.txt file. I think we could move it into the root now?

Makes sense, I'll do that in the `only-poetry` branch and then update here as well.

legoktm (commented):
Not giving it the green check yet since it's still a draft and the sd-proxy side needs to be finished, but overall this LGTM :)

scripts/update-requirements
```diff
@@ -31,20 +38,43 @@ def main():
...
+    else:
+        dependencies = utils.get_requirements_names_and_versions(requirements_file)
+
+    dependencies = [dep for dep in dependencies if dep[0] not in SKIP_PACKAGES]
 
     verify_sha256sums_file(project)
 
+    # If there is no 'requirements' directory, we look in the project root
+    if not (pkg_dir / args.requirements).exists():
+        args.requirements = ""
```
legoktm, line 71: This feels a bit fragile and it would be nice to avoid mutating `args`.

Maybe something like:
```python
if (pkg_dir / args.requirements).exists():
    build_requirements_file = pkg_dir / args.requirements / "build-requirements.txt"
else:
    # If there is no 'requirements' directory, we look in the project root
    build_requirements_file = pkg_dir / "build-requirements.txt"
```

scripts/utils.py
```diff
@@ -2,19 +2,176 @@
...
+) -> list[str]:
+    """
+    Identify main (non-development) requirements. poetry.lock does not preserve
+    the distinction between different dependency groups, so we have to parse
+    both files.
+    """
+    pyproject_dict = tomllib.loads(path_to_pyproject_toml.read_text())
+
+    # Extract main dependencies
+    main_dependencies = list(pyproject_dict["tool"]["poetry"]["dependencies"].keys())
```
legoktm, line 81: nit: You don't need `.keys()`, running `list()` over a dict will only return keys.

scripts/utils.py
```diff
@@ -2,19 +2,176 @@
...
+    # Create a set to keep track of main and transitive dependencies (set ensures
+    # no duplicates)
+    relevant_dependencies = set(main_dependencies)
+
+    # Identify transitive dependencies (may be enumerated in lockfile
+    # before the dependency which declares them)
+    for package in parsed_toml.get("package", []):
+        package_name = package["name"]
+        if package_name in main_dependencies:
+            for sub_dependency in package.get("dependencies", {}).keys():
```
legoktm, line 98: nit: Don't need `.keys()`, iterating over a dict already just iterates over keys.

scripts/verify-hashes
```diff
@@ -3,96 +3,69 @@
...
+Ensure that we have at least one signed SHA256 sum for each dependency declared
+in a project's requirements file. This will be the source tarball.
+"""
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("Usage: ./scripts/verify-hashes path/to/sha256sums.txt")
         sys.exit(1)
 
-sys.exit(0)
+    sha256sum_data = {}
```
legoktm, line 33: Can you put all this code into a main() function instead of having it be file level? Keeping it above would also reduce the diff and make it easier to track which code is new and which isn't.

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
...
+def test_get_relevant_poetry_dependencies():
+    output = get_relevant_poetry_dependencies(PYPROJECT_TOML_PATH, POETRY_LOCK_PATH)
+    # Ensure we correctly ignore development-only dependencies
+    assert "pytest" not in output
+    assert sorted(output) == sorted(EXPECTED_DEPENDENCY_NAMES)
+
+
+def _check_hashes(output):
+    assert sorted(list(output)) == sorted(EXPECTED_KEYS)
+    for dep in output.keys():
```
legoktm, line 57: nit: No need for `.keys()`. I'd actually suggest `.items()` so you don't need to index back into the dict with `output[dep]`.

legoktm (changes requested):
Code overall looks good, just some small things. Running through the test plan now.

scripts/utils.py
```diff
@@ -2,19 +2,176 @@
 Shared functions between various scripts
 """
 import re
+import sys
+
+try:
+    import tomllib
+except ModuleNotFoundError:
+    # pre-Python 3.11 compatibility
+    import toml as tomllib
```
legoktm, line 11: This will fail in existing CI bullseye pipelines because we don't activate the virtualenv: https://github.com/freedomofpress/securedrop-log/blob/487707a028b19931b013f0ea66d953fb1af20914/.circleci/config.yml#L32

So either we could go around and add `source .venv/bin/activate` to all the CI pipelines before merging this, or we could also catch this import, set tomllib to False and e.g. `print("tomllib/toml not found, poetry support unavailable")`

scripts/verify-hashes
```diff
@@ -3,96 +3,69 @@
...
-#   'alembic': ['04bcb970ca8659c3607ddd8ffd86cc9d6a99661c9bc590955e8813c66bfa582b']
-# }
+from pathlib import Path
+import utils
+
+try:
+    import tomllib
+except ModuleNotFoundError:
+    # pre-Python 3.11 compatibility
+    import toml as tomllib
```
legoktm, line 13: Same issue here. Maybe it would be easier if we did `from utils import tomllib` instead of reimplementing the try/catch.

legoktm:
I was able to successfully run through the whole test plan on bullseye (bookworm is busted because of the pyyaml wheels issue). I think the main issue is figuring out how we want to handle back-compat of CI pipelines that currently don't activate the venv - either have them activate the venv before this lands or write in extra back-compat in this PR so it gracefully works.

Regarding the:
```diff
-REQUIREMENTS_FILE=requirements/build-requirements.txt
+REQUIREMENTS_FILE=build-requirements.txt
```

change, I'm wondering if you want to just have that be a PR that we coordinate merging at the same time, or write some temporary if exists/else logic in the make file to detect the correct path that can be removed post-poetryification.

Other than that, I think we're very close!!

scripts/utils.py
eloquence: We're modifying the keys as well; I've changed the code to do that on the dict instead.

scripts/verify-hashes
eloquence: Turns out we're not even using the import here, it all got moved into utils :sweat_smile:. Removed.

scripts/utils.py
eloquence: I think I like the second solution better, will poke.

scripts/verify-hashes
eloquence: Done in 4d11da00c759d2918a2f9bb63a448fccda50b978

scripts/utils.py
eloquence: OK, this is done minimally. As far as I can tell, pipelines should still succeed against packages that don't use Poetry. The `tomllib` name is only used inside the Poetry-related functions and there should be no `NameError` during a run against a non-poetry package.

eloquence:
Thanks @legoktm, I think I've addressed your comments, but let me know if not! I've opened a new draft PR with just the `build-requirements.txt` change, so we can merge this together with the `securedrop-proxy` changes after this main compat PR lands.

legoktm (approved):
Overall looks good to me - do you want to squash or merge as-is?

eloquence:
Maximally squashed :)

PR #469, opened 2023-10-25
Update build-requirements.txt path for securedrop-proxy package

eloquence:
This should be merged together with https://github.com/freedomofpress/securedrop-proxy/pull/122 and will fail until then.

### DIFF
--- a/securedrop-proxy/debian/rules
+++ b/securedrop-proxy/debian/rules
@@ -3,7 +3,8 @@
 %:
 	dh $@ --with python-virtualenv
 
-REQUIREMENTS_FILE=requirements/build-requirements.txt
+# For backwards compatibility; can be set to build-requirements.txt once fully switched to new path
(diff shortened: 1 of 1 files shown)


securedrop-proxy/debian/rules
```diff
@@ -3,7 +3,7 @@
 %:
 	dh $@ --with python-virtualenv
 
-REQUIREMENTS_FILE=requirements/build-requirements.txt
+REQUIREMENTS_FILE=build-requirements.txt
```
legoktm, line 6: What do you think of something like:
```
REQUIREMENTS_FILE=$(shell test -e build-requirements.txt && echo "build-requirements.txt" || echo "requirements/build-requirements.txt")
```

That should support both the old and new locations. I would much prefer being able to rely on CI passing at every stage of this if possible...

eloquence: Yep, that works well. I'd prefer to revert that pretty quickly so we don't forget about it, will get a PR ready to do so.

legoktm (approved):

PR #472, opened 2023-11-03
Add backwards-compatible reference to new build-requirements.txt

legoktm:
This file will be moving to the repository root in the near future as part of our migration to poetry, making the packaging compatible with both eases the transition.


rocodes (approved):
:+1:

PR #474, opened 2023-11-13
Fix issues around case-sensitive matching in poetry support

legoktm:
poetry doesn't normalize case in package names, so in securedrop-client we have jinja2 depending on "MarkupSafe", but the package itself in the lock file is named "markupsafe".

We can avoid this by changing the relevant_dependencies to all be lowercase when reading them out of pyproject.toml (the "main" dependencies) and when reading from poetry.lock.

## Test plan

* [x] Running `PKG_DIR=../securedrop-proxy make requirements` results in no changes
* [x] repeat for sd-export
* [x] and for sd-log
* [x] and for sd-client with https://github.com/freedomofpress/securedrop-client/pull/1671 checked out

### DIFF
--- a/scripts/utils.py
+++ b/scripts/utils.py
@@ -80,26 +80,25 @@ def get_relevant_poetry_dependencies(
     """
     pyproject_dict = tomllib.loads(path_to_pyproject_toml.read_text())
 
-    # Extract main dependencies
-    main_dependencies = pyproject_dict["tool"]["poetry"]["dependencies"]
-
-    # Remove 'python' as it's not a package dependency
-    if "python" in main_dependencies:
-        del main_dependencies["python"]
-
-    parsed_toml = tomllib.loads(path_to_poetry_lock.read_text())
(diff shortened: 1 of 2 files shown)


scripts/utils.py
```diff
@@ -80,28 +80,27 @@ def get_relevant_poetry_dependencies(
...
     for package in parsed_toml.get("package", []):
         package_name = package["name"]
-        if package_name in main_dependencies:
+        if package_name in relevant_dependencies:
             for sub_dependency in package.get("dependencies", {}):
-                relevant_dependencies.add(sub_dependency)
+                relevant_dependencies.add(sub_dependency.lower())
 
-    return list(relevant_dependencies)
+    return [dep for dep in relevant_dependencies]
```
eloquence, line 103: Out of curiosity, what's the benefit of avoiding list() here?

eloquence (approved):
LGTM!

PR #475, opened 2023-11-30
Cleanup old RPM nightlies too

### LINKED ISSUE #435: Automate cleanup of old rpm nightlies
Now that we are generating rpm nightlies, we also need a cleanup script for them :)

legoktm:
## Status

Ready for review, but needs https://github.com/freedomofpress/securedrop-yum-test/pull/51 to be merged first.

## Description

Extend the `clean-old-packages` script so it works for RPMs too. We use the Python RPM bindings to extract package name and version and then shell out to `rpmdev-vercmp` to sort them.

The behavior is now the same as the deb cleanup, when new nightly packages are pushed, it'll clean up old nightlies in the same commit. RC packages (in the `f32` component) will be cleaned up in a separate commit, if necessary.

Fixes #435.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ common-steps:
       name: Commit workstation debs for deployment to apt-test.freedom.press
       command: |
         apt-get update
-        apt-get install -y ca-certificates git git-lfs openssh-client python3 python3-debian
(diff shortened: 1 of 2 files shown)


cfm (approved):
Looks great, @legoktm.  Thank you!

PR #480, opened 2023-12-13
Normalize names per PEP503; update test fixtures

### LINKED ISSUE #479: Poetry logic does not fully account for underscores in package names
Even with #474, our current `poetry` logic is still susceptible to inconsistencies in how dependencies may appear in `pyproject.toml` and `poetry.lock`.

## Steps to reproduce

1. In a Poetry-managed project, declare a dependency `flit_core` in `pyproject.toml`
2. Update the lockfile
3. Attempt to build the wheels

## Expected behavior

New dependency will be picked up

## Actual behavior

`flit-core` (which is how it appears in `poetry.lock`) is not picked up

eloquence:
Fixes #479 

`pyproject.toml` allows dependency names to be specified in all permissible variations before they get normalized per [PEP 503](https://peps.python.org/pep-0503/#normalized-names) in the lockfile. This wasn't previously taken into account when parsing requirements. (Dependency names can also appear inside of quotation marks, but that's already normalized when parsing the TOML.)

In principle, we shouldn't need to normalize the names we encounter in `poetry.lock` itself, but since the format is not supposed to be considered stable, we're erring on the side of caution.

## Test plan

- Follow the STR in #479
- [x] Confirm that the issue is fixed


scripts/utils.py
```diff
@@ -177,3 +177,12 @@ def get_requirements_from_poetry(
         requirements.append(requirement_line)
 
     return "\n".join(requirements)
+
+
+def normalize(dependency_name: str) -> str:
+    """
+    For consistent comparisons between pyproject.toml and lockfile,
+    normalize names consistent with PEP503:
+    https://peps.python.org/pep-0503/#normalized-names
```
cfm, line 186: Nit: I found it helpful to understand that this is a blessed, self-sufficient implementation of this normalization:

```suggestion
    normalize names consistent with PEP503.  This implementation is drawn directly from
    <https://peps.python.org/pep-0503/#normalized-names>.
```

cfm (approved):
Looks good to me!  I've left one nit inline for your consideration, but I'll be happy to merge with or without it.

cfm (approved):
Thanks, @eloquence!

PR #483, opened 2024-01-10
Test that workstation-bootstrap wheels are reproducible too

legoktm:


tests/test_reproducible_wheels.py
```diff
@@ -22,3 +22,13 @@ def test_wheel_builds_match_version_control(name):
...
     # in any way, so mostly useful in CI.
     assert subprocess.check_output("git status --porcelain".split()) == b""
+
+
+def test_workstation_bootstrap_wheels():
+    subprocess.check_call([
+        "./scripts/build-sync-wheels", "--pkg-dir", "./workstation-bootstrap",
+        "--project", "workstation-bootstrap"])
+    # Check for modified files (won't catch new, untracked files)
+    subprocess.check_call(["git", "diff", "--exit-code"])
```
eloquence, line 32: In #484 I've followed the linter suggestion to prefix with `/usr/bin/` instead of silencing the linter; if you agree that's generally good practice might be nice to do so here as well. Same with subsequent `git` command.

zenmonkeykstop:
blocked on https://github.com/freedomofpress/securedrop-builder/pull/484

legoktm:
Rebased and updated with ruff fixes and formatting.

PR #484, opened 2024-01-10
Add ruff for lint & formatting checks; apply some suggested fixes

### LINKED ISSUE #463: Add ruff, black to Makefile and CI
We should run linter and code formatter checks against the Python code in `scripts/` and `tests/`.

eloquence:
Adds a GitHub action that runs format & lint checks.

Draws from https://github.com/freedomofpress/securedrop/pull/6961 for the ruff config.

I've implemented several of the linter suggestions while ignoring others.

This PR also applies `ruff`'s code formatting (which in turn is a drop-in replacement for `black`). `ruff` appears to be a bit less aggressive than `black` in enforcing line length, so we have to disable the line length check itself.

NB: Our undocumented use of `test-requirements.txt` is another potential candidate for Poetry; I've saved this for a follow-up change.

Fixes #463 

## Testing

- Visual review
- Review output of new "checks" GitHub actions
- Install `test-requirements.txt` in a venv and verify that Makefile targets work as expected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,6 @@ on: [push, pull_request]
 jobs:
   checks:
     runs-on: ubuntu-latest
-    # We use a standard Debian image to mirror a typical developer environment.
-    # This should be updated whenever a new Debian stable version is available.
     container: debian:bullseye
     steps:
       - uses: actions/checkout@v4
@@ -15,9 +13,6 @@ jobs:
         run: |
           apt-get update && apt-get install --yes --no-install-recommends make python3-pip
           pip install -r test-requirements.txt
-      - name: Run lint check
+      - name: Run lint checks
         run: |
-          make check-lint
-      - name: Run format check
-        run: |
-          make check-format
+          make lint
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+---
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    # We use a standard Debian image to mirror a typical developer environment.
+    # This should be updated whenever a new Debian stable version is available.
+    container: debian:bullseye
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          apt-get update && apt-get install --yes --no-install-recommends make python3-pip
+          pip install -r test-requirements.txt
+      - name: Run lint check
+        run: |
+          make check-lint
+      - name: Run format check
+        run: |
+          make check-format
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,22 @@
 DEFAULT_GOAL: help
 SHELL := /bin/bash
 
+.PHONY: check-lint
+check-lint:
+	@ruff check .
+
+.PHONY: fix-lint
+fix-lint:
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -1,6 +1,22 @@
 DEFAULT_GOAL: help
 SHELL := /bin/bash
 
+.PHONY: check-lint
+check-lint:
```
legoktm, line 5: Do we need 4 separate make targets? I think we just need one "make lint" (this seems standard in most of our other repos) that does the linting and "make fix" that does all the fixing (c.f. https://github.com/freedomofpress/securedrop-tooling/issues/11).

pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+]
+ignore = [
+    # Find contextlib.suppress() is harder to read
+    "SIM105",
+    # Find ternary statements harder to read
+    "SIM108",
+    # Flags any subprocess use
+    "S603",
+    # Ruff formatter makes a best effort, but may result in some lines exceeding 100 characters
+    "E501",
```
legoktm, line 46: How bad is it? Is it unreasonable to annotate the few exceptions with `noqa` or are there too many? (If it's that bad maybe we don't adopt ruff formatting just yet?)

I guess line length doesn't matter as a lint since if the formatting doesn't match it'll still fail CI, but it feels weird to me to set a max line length and then also ignore it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
+---
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    # We use a standard Debian image to mirror a typical developer environment.
+    # This should be updated whenever a new Debian stable version is available.
```
legoktm, line 10: I don't think this is correct since bookworm is available, it's mostly when we move SDW over to the new version. I don't think we need to annotate our choice, we haven't in most other CI manifests.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
+---
```
legoktm, line 1: Since this is the first GHA workflow, can you please also add the dependabot config (https://github.com/freedomofpress/securedrop/pull/7102)

legoktm (changes requested):
Thanks for venturing into the brave new world of `ruff format` :) Left some notes but overall I'm excited

.github/workflows/ci.yml
eloquence: Done!

.github/workflows/ci.yml
eloquence: Fair enough; we have the same comment in the docs repo (https://github.com/freedomofpress/securedrop-docs/blob/main/.github/workflows/ci.yml) so probably should remove it there as well.

pyproject.toml
eloquence: It was just one file. I've removed the exception and fixed up that one file; we just need to be aware that line length will sometimes require a bit more manual futzing than with `black`.

Makefile
eloquence: Happy to combine them, done.

pyproject.toml
legoktm: Thanks, that seems fine with me, if it is problematic I think we'll only find out in much larger repos.

legoktm (approved):

PR #485, opened 2024-01-10
Unbreak nightlies failing due to CircleCI behavior change

eloquence:
The `pipeline.schedule.name` variable now seems to be unset, causing nightlies to fail. We were not actually using `SCHEDULE_NAME` for anything, so this should be otherwise a noop.


legoktm (approved):
Wonderful, thank you :)

PR #490, opened 2024-02-06
Remove files related to Debian packaging, now in client repo

legoktm:
All of the packaging logic has moved to the securedrop-client repository, leaving this responsible for the localwheels building. (RPM nightlies will be moved out at a later time.)

Nearly everything has been migrated, the things that haven't (lint-desktop-files, test_deb_package.py and clean-old-packages) have open issues/PRs for them.

One explicit workflow change is that PRs to this repo will not test
package builds, because doing so causes a catch-22 scenario when the
build-requirements.txt in the client repo hasn't been updated yet, so it
fails.


securedrop-export/debian/securedrop-export.triggers
rocodes, line 1: I noticed this file isn't present in `securedrop-client` after the refactor changes. Is that intentional?

securedrop-workstation-viewer/debian/control
rocodes, line 1: Same as previous - I may be missing something about the refactor but are we missing some workstation-viewer package config in the client repo now?

rocodes (commented):
Thank you @legoktm for all the work :) Left a couple clarification questions re: changes between here and the client repo (around the .triggers files and the workstation-viewer package). Otherwise LGTM

securedrop-export/debian/securedrop-export.triggers
legoktm: Yep, we are no longer using dh-virtualenv so there's no trigger for us to register on to. (And the triggers file was mostly useless too, c.f. https://github.com/freedomofpress/securedrop/pull/6231)

securedrop-workstation-viewer/debian/control
legoktm: The workstation-viewer package is just a "metapackage", it doesn't contain any files of its own, just depends on other packages, so it's fully contained in the d/control file now: https://github.com/freedomofpress/securedrop-client/blob/main/debian/control#L47

best way to verify this is build the new debs in the client, grab an old deb from apt-prod and diffoscope them :)

rocodes (approved):
Thanks for clarifications @legoktm - lvgtm

PR #494, opened 2024-03-27
Update ruff config format; apply latest fix

eloquence:
We're currently installing the latest version of `ruff` in CI without pinning. Until we change that, we'll have to keep up with upstream behavior changes. This PR does that - it updates the configuration format to silence a couple of deprecation warnings, and applies a small formatting change that a previous version did not catch.


legoktm (approved):
We should probably also pin it properly...

PR #496, opened 2024-04-12
Migrate tests to GitHub Actions

eloquence:
Migrates `make test` and `make reprotest` tests to GitHub Actions and removes the CircleCI configuration.

Fixes https://github.com/freedomofpress/securedrop-builder/issues/482.

## Test plan

- Visual review of configuration
- Inspect test results to ensure same tests are running as previously


eloquence:
This is ready for review, but does not include `make reprotest` yet -- the test is runnable, but it failed on https://github.com/freedomofpress/securedrop-builder/actions/runs/8745779791/job/24001431722. I'd like to determine if that's an expected failure or if it's not running correctly for some reason. (And yeah, then we can add a nightly as well.) 

On the other hand, the CircleCI config appears to be borked on this repo -- CircleCI is complaining about lack of SSH checkout key. I defer to SD if that's worth fixing before we've got the job switched over. If not, I can poke more at it Monday/Tuesday to finish things up.

eloquence:
Re: reprotest, I do see a lot of `warning: Not a git repository. Use --no-index to compare two paths outside a working tree`, so it does not look like an expected test failure but potentially a need to adapt the checkout logic in that test flow.

eloquence:
`reprotest` is passing now. :tada: The issue was twofold:
- We need to set git's `safe.directory` option for the `git` operations to succeed
- As per CircleCI, we need to avoid running tests as root due to `tar` reproducibility issues when run as root

This PR is also removing the CircleCI config, so CircleCI is complaining now, but if the rest of it looks good, we can remove the CircleCI config fully from this repo prior to merge.

.circleci/config.yml
```diff
@@ -1,69 +0,0 @@
...
-    working_directory: "/srv"
-    steps:
-      - checkout
-      - *installdeps
-      - *installtestdeps
-      - run:
-          name: install test requirements and run tests
-          command: |
-            adduser --system ci --ingroup root
-            sed -i -re "292s/^(\s+).*\$/\1return _.prepend_to_build_command_raw('')/" /usr/lib/python3/dist-packages/reprotest/build.py
```
eloquence, line 49: This in-place patching goes back to 18770bd44ec54da86f436515da4452c05111605e and, if it was still relevant at all, was not necessary for the wheel reproducibility checks, which do not use the `reprotest` package that is being patched here. The actual Debian reproducibility check has been migrated into `securedrop-client` and no longer uses `reprotest`.

legoktm (approved):
Awesome, LGTM! We will want to move these jobs to bookworm ASAP now (c.f. https://github.com/freedomofpress/securedrop-client/pull/1958) but we can do that in a separate PR. Will poke infra for the CircleCI disconnect.

PR #498, opened 2024-04-24
Verify sha256sums.txt and signatures in CI

legoktm:
Avoids a situation like <https://github.com/freedomofpress/securedrop-builder/pull/497#issuecomment-2073826834>, where wheels are updated, but the sha256sums and associated signatures are not.

## Test plan
* [ ] visual review
* [ ] CI passes


rocodes (approved):
## Test plan
* [x] visual review
* [x] CI passes

LGTM - thank you for preparing and @eloquence thanks for the prompt

PR #499, opened 2024-05-24
Remove proxy wheels, no longer used

legoktm:
## Status

Ready for review

## Description

As of proxy v2, it's now implemented in Rust so there is no need for these Python wheels anymore.

## Test plan

* [x] Visual review
* [x] CI passes


cfm (approved):
Thanks for cleaning this up, @legoktm.  I note for the record that this also removes _checking_ the reproducibility of `securedrop-proxy`'s wheels.

PR #501, opened 2024-07-22
Upgrade workstation-bootstrap packages

### LINKED ISSUE #500: Upgrade setuptools for CVE-2024-6345
Related: https://github.com/freedomofpress/securedrop/pull/7205

Upgrading setuptools is turning into cascading upgrades for everything, we also need to upgrade pip, and if you don't upgrade packaging, you'll hit https://github.com/pypa/setuptools/issues/4483. But packaging needs a newer version of flit-core, which triggers conflicts with pep517, so let's also upgrade build to drop that dependency.

legoktm:
Motivated by CVE-2024-6345 in `setuptools`. But just upgrading `setuptools` forces upgrades of `pip`, which in turn requires a newer `packaging`, which then needs a newer `flit-core`. But that is then incompatible with `pep517`, so upgrade `build` to drop that dependency.

Upgraded packages:
* setuptools 54.0.0 -> 70.3.0
* setuptools-scm 5.0.2 -> 8.1.0
* pip 21.3.1 -> 24.1.2
* packaging 20.9 -> 24.1
* flit-core 2.3.0 -> 3.9.0
* build 0.3.0 -> 1.2.1
  * colorama is skipped, it isn't hard dependency and only needed for Windows anyways
  * pep517 0.9.1 -> pyproject_hooks 1.1.0 (package was kind of renamed)

Dropped dependencies:
* pyparsing
* pytoml
* toml

Fixes #500.

## Diff reviews
Assuming all the diff reviews from https://github.com/freedomofpress/securedrop/pull/7205 are completed, we additionally need to review:
* [ ] pip 21.3.1 -> 24.1.2 
* [x] flit-core 2.3.0 -> 3.9.0 https://github.com/freedomofpress/securedrop-builder/wiki/flit_core-2.3.0-to-3.9.0
* [x] build 0.3.0 -> 1.2.1  https://github.com/freedomofpress/securedrop-builder/wiki/build-0.3.0-to-1.2.1
* [x] pep517 0.9.1 -> pyproject_hooks 1.1.0 (package was kind of renamed): https://github.com/freedomofpress/securedrop-builder/wiki/pep517-0.9.1-to-pyproject_hooks-1.1.0


rocodes:
I will start tomorrow with the bottom of this list (`pyproject_hooks`) and work upwards, but may only get through 1 tomorrow depending on support load (and size of diff).

legoktm:
Discussion during last week's standup (or team meeting, I don't remember), was that reviewing pip (the last remaining review needed) is not worth it because of how large it is and how central it is to Python. @zenmonkeykstop was going to do some more historical research and put together a proposal. Now that we have most/all of the team back, I'll do a vibe check in standup if we're okay not reviewing pip and moving forward with this before the formal proposal.

legoktm:
Consensus today is that we're OK not having a diff review for pip to land this, so this is now ready for review.

cfm (approved):
Thanks for the detailed write-up, @legoktm.  I've confirmed that `make reprotest` passes locally.

PR #506, opened 2025-01-03
Specify test dependencies using poetry

### LINKED ISSUE #466: Add Makefile target for test reqs (or include in install-deps)
Currently `test-requirements.txt` is sitting there, undocumented, waiting for the developer to discover and install it as needed when `make test` fails. Let's give it a nice comfortable home somewhere, be it as its own Makefile target, or just included in `install-deps` if we don't mind bundling it up with the build prerequisites.

legoktm:
So they are pinned and much easier to discover.

Fixes #466.


cfm (approved):
Thanks, @legoktm.

PR #507, opened 2025-01-03
Lint our GitHub Actions workflows with zizmor

legoktm:
_**Note**: this depends on https://github.com/freedomofpress/securedrop-builder/pull/506, which should be merged first_

We just need to set `persist-credentials: false` in all of our workflows.

Refs <freedomofpress/securedrop-tooling#18>.


cfm (approved):
Approving after rebasing from `main` after #506.  This follows the precedent established in other pull requests towards freedomofpress/securedrop-tooling#18 and runs as expected.

PR #508, opened 2025-01-30
chore: update to `pip` 25.0 for Safety 75180

cfm:
This does not require diff review, per <https://github.com/freedomofpress/securedrop-dev-docs/blob/4a42bbb4eb4007edcccb317da5f19998dc3175e0/docs/dependency_updates.rst?plain=1#L52-L53>.


workstation-bootstrap/sha256sums.txt
```diff
@@ -21,6 +21,8 @@ f266816461e421e2bbdb61e6e122d01fe638b5710698879ab283cc456aabd1d1  pip-21.0.1-py3
 fd11ba3d0fdb4c07fbc5ecbba0b1b719809420f25038f8ee3cd913d3faa3033a  pip-21.3.1.tar.gz
 45941b49b443d67e8f2c820b4fa9d4aa916326f897abfa1068914f1b0ebc6eb2  pip-24.1.2-py3-none-any.whl
 e5458a0b89f2755e0ee8c0c77613fe5273e05f337907874d64f13171a898a7ff  pip-24.1.2.tar.gz
```
legoktm, line 23: Should these old pip lines also have been deleted?

legoktm (commented):
One question for you @cfm. I was going to say that we needed to leave the N-1 package in place so it didn't break client builds, but this is just in the bootstrap so it's not pinned in the client and it produces identical wheels to what we currently have; wonderful.

workstation-bootstrap/sha256sums.txt
cfm: Good catch.  See the corrected 2ea2da8b226c17346f82e4c600c81b6878286341 and 9dcf3826ee8dea60b2cec52af4363b0f7110879c for the missing enforcement.

legoktm (approved):
Good find on the missing verification, thanks!

PR #509, opened 2025-02-11
Remove Python 3.9 wheels that were for bullseye

legoktm:
Let's clean these up before we add trixie wheels.


legoktm:
Please note that https://github.com/freedomofpress/securedrop-client/pull/2386 should be merged after this.

PR #511, opened 2025-03-20
chore: `jinja2` 3.1.6

cfm:
Towards freedomofpress/securedrop-client#2410.

diff review: https://github.com/freedomofpress/securedrop-builder/wiki/jinja2-3.1.3-to-3.1.6


legoktm (approved):
LGTM, spent a while trying to find out why "jinja2" is now lowercase, but it's the same way on pypi too, starting with 3.1.3. Hopefully that won't cause any issues with our tooling...

Issue #464, opened 2023-09-13
Update bootstrap requirements


eloquence:
Our bootstrap requirements are very outdated, making it difficult to add any current dependencies without running into dependency hell.

- `build` is at version 0.3.0 from February 2021
- `pip` is at 21.3.1 from October 2021
- `wheel` is at 0.36.2 from December 2020

And so on. In the Python ecosystem, build/dependency issues proliferate quickly as you fall behind, due to the huge variety of different systems involved, and the conflicting version constraints between them. If we'll want our build environment to be maintainable in this form, we'll need to update it more regularly.

eloquence:
One question this raises for me is if the utility of local wheels (intended to guard against compile-time security risks from upstream) really holds up to scrutiny when compared to the effort required to actually maintain them -- and the cost of falling behind, including in patching potential security issues in the build requirements.

legoktm:
> if the utility of local wheels (intended to guard against compile-time security risks from upstream) really holds up to scrutiny

I roughly had the same opinion when I started dealing with all of this but I now think it's totally worth it. It's insulated us from a number of breaking changes in the Python ecosystem (most recently the PyYAML + Cython incompatibility) and adds security benefits like having cryptography dynamically link against OpenSSL instead of using the upstream wheels which bundle it.

I've yet to personally run into issues with the bootstrap being that outdated but yeah, it's probably good to update it. I'll comment on the other issue though on whether we need poetry in the bootstrap or not.

legoktm:
> when compared to the effort required to actually maintain them

...and on the flip side this is way too manual of a process, it really should be as simple as updating the poetry/requirements input files, running a command that takes care of all the build stuff in containers, and then a final step to sign things, not 5-6 intermediate steps :/

legoktm:
https://github.com/freedomofpress/securedrop-builder/pull/501 upgraded pip and build but not wheel. wheel is under PyPA and doesn't need a diff review so hopefully can be upgraded pretty easily.

PR #512, opened 2025-04-21
chore: update to `setuptools` >= 78.1.1 for Safety 76752

cfm:
This does not require diff review, per <https://github.com/freedomofpress/securedrop-dev-docs/blob/4a42bbb4eb4007edcccb317da5f19998dc3175e0/docs/dependency_updates.rst?plain=1#L52-L53>.


cfm:
However, `reprotest` suggests that all of our other wheels will have to be rebuilt using the new `setuptools` in `workstation-bootstrap`.  I'll come back to that tomorrow.

cfm:
> * wheels are now lowercase (SQLAlchemy -> sqlalchemy). Shouldn't we also remove the old wheels?

Yes, we've been seeing this normalization as we update things incrementally.  We don't seem to be in the habit of removing old wheels when this or other formats change:

```sh-session
user@sd:~/securedrop-builder/securedrop-client/wheels$ git log -1 --oneline --no-show-signature
d7ee80c (HEAD -> safety-76752, origin/safety-76752) chore: update wheels with new setuptools
user@sd:~/securedrop-builder/securedrop-client/wheels$ for x in $(ls {M,m}ako-1.2.2*); do ls -al $x; git log -1 --oneline --no-show-signature $x; done
-rw-r--r-- 1 user user 78693 Apr 22 12:41 mako-1.2.2-py3-none-any.whl
d7ee80c (HEAD -> safety-76752, origin/safety-76752) chore: update wheels with new setuptools
-rw-r--r-- 1 user user 78664 Aug 20  2024 Mako-1.2.2-py3-none-any.whl
be99501 Rebuild all wheels with upgraded bootstrap packages
-rw-r--r-- 1 user user 490741 Apr 22 12:41 Mako-1.2.2.tar.gz
40f842d Add mako 1.2.2 wheel
```

I'll drop a note in #464 that this would be a good thing to do the next time we update and rebuild dependencies across the board.


> * The only potentially-impacting changes noticed by diffoscope are changes in `Requries-dist` (some removals and some additions). The [docs](https://packaging.python.org/en/latest/specifications/core-metadata/#requires-dist-multiple-use) are not very clear on what impact this has or why it exists. @cfm do you know if this has any impact?

I had to read up on it too, but it looks like the combination of `Provides-Extra` and `Requires-Dist` is similar to Cargo features in Rust: a way to specify optional features (`Provides-Extra`) of the package that add further dependencies (`Requires-Dist`).  So we're seeing changes in these packages' metadata, but the wheels and signed hashes here are still definitive for what we're actually installing.

PR #513, opened 2025-05-29
Update poetry to 2.1.3

rocodes:
WIP:  refs https://github.com/freedomofpress/securedrop-client/pull/2461, needs CI fixups, I think needs to be merged first/in coordination


legoktm:
The changes look good, but I think we need to fix #514 first, as right now `tests/test_reproducible_wheels.py::test_workstation_bootstrap_wheels` is failing, and that's what you're modifying. I'm 99% sure that's a side-effect of the other one being busted, but want to be sure.

rocodes:
yup, agreed, and thank you for taking a look - I put up https://github.com/freedomofpress/securedrop-builder/pull/518, I think we're getting closer to untangling it all :)

legoktm (changes requested):
I see you didn't end up making any changes to `workstation-bootstrap/pyproject.toml`, and I verified that running `poetry lock` works fine in that directory and matches what you committed. But in https://github.com/freedomofpress/securedrop-client/pull/2461/ you switched to using the `project.dependencies` style. I think we should be consistent on that one way or another.

I primarily bring it up because in utils.py, the `get_relevant_poetry_dependencies` function reads the `tools.poetry.dependencies` field, which is why the client PR is failing the verify-builder-sync job (https://github.com/freedomofpress/securedrop-client/actions/runs/15398271213/job/43324390552?pr=2461). So I think either we stick with the old tools.poetry.dependencies style, or need a small fix in utils.py to support the new style?

legoktm:
Oooh TIL about `poetry check`. That seems like something that might be useful for us to add in CI as a lint step (but as a later thing, not now)

rocodes:
(Addressed utils.py parsing and CI and squashed component poetry updates into a single commit)

PR #519, opened 2025-06-24
Add .github/CODEOWNERS

legoktm:
For our branch protection rules.


zenmonkeykstop:
Approving as is, #520 fixes the expired key.

PR #522, opened 2025-07-08
Update rowen.pub

rocodes:
(updates my pubkey)

see https://keys.openpgp.org/vks/v1/by-fingerprint/91B3D67D8758449546256E7EFC72AA838C444B47


legoktm (approved):
Confirmed that both the old and new key files have the same fingerprint: 91B3D67D8758449546256E7EFC72AA838C444B47.

PR #529, opened 2026-04-16
Update pyproject.toml and poetry.lock with Poetry 2.3.3

mdaleo-fpf:
Works towards: https://github.com/freedomofpress/infrastructure/issues/6688

Small change to point at a minimum Poetry version of 2.3.3


legoktm (changes requested):
Thanks, the CI workflows also need to be updated to upgrade poetry

mdaleo-fpf:
@legoktm It should be good now.

legoktm (approved):
Thanks!

